PR #2, opened 2015-01-05
Refs #8432 - Allow either foreman.url or proxy.url to be used

GregSutcliffe:
I don't see any reason not to support both foreman.url and proxy.url, especially given how trivial it is, and that existing users will forget to update their config. Here's the patch, thoughts @lzap?

### DIFF
--- a/root/usr/bin/discover-host
+++ b/root/usr/bin/discover-host
@@ -58,7 +58,7 @@ def discover_server
 end
 
 def discover_by_url
-  url = cmdline 'proxy.url'
+  url = cmdline('proxy.url') || cmdline('foreman.url')
   log_msg "Discovered by URL: #{url}" if url
   URI.parse(url)
 rescue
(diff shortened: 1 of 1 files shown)


lzap:
Since we error out hard when you dont specify `proxy.type` I don't see much added value here and I'd rather be very explicit. But I am okay with this backward compatibility mode. Any objections @shlomizadok ?

GregSutcliffe:
@lzap i don't see any hard error for missing `proxy.type` - it works fine for me without it, and defaults to contacting foreman directly.

GregSutcliffe:
I've also updated the error message for the missing parameter to inform the user of what's missing

lzap:
Ok this makes sense, thanks.

GregSutcliffe:
@lzap actually, I was wrong, it defaults to the proxy type (although there is no hard error either way). I think we should default to type=foreman for one release, since it's easy to do and doesn't cause issues with legacy configurations.

Assuming you agree, I've added the change of default to the PR.

lzap:
Tested for both cases, looks good. Thanks. bfaf660

GregSutcliffe:
Thanks @lzap - We should start keeping release notes for 2.0, so that we can tell people to update the config to proxy.type = foreman if they're not going to deploy the proxy plugin :)

PR #3, opened 2015-01-06
Removing dracut patch

lzap:
DO NOT MERGE I need to test this against RHEL7 (CentOS7 is fine)


lzap:
Ok this is ready for merge @stbenjam it works fine. Downstream we build in brew therefore `build-livecd-root` script is not used at all. My image built without this hack works just fine, otherwise it would not boot (some older Fedoras don't).

stbenjam:
Merged as 1366710ac809ca1a70e5cd0c2c72fdc46c6cbba1, thanks @lzap !

PR #4, opened 2015-01-07
Refs ??? -  WIP Extend image at runtime via a zipfile

GregSutcliffe:
Current WIP. Working:

* Multiple zips
* Custom facts
* Autostart script(s)
* Binaries
* LD load path
* RUBYLIB path
* Refactoring

Todo:

* download location is currently hardcoded - add routes to the proxy/foreman plugins and a PXE option to specify the path
* example zip generator

@lzap thoughts so far?


README
```diff
@@ -53,11 +53,33 @@ You can also use the image standalone (without TFTP under Foreman's
...
+`bin` is added to PATH, you can place binaries here. If this directory
+contains an executable `autostart` script, it will be run *before* the
+image is registered to Foreman. `facts` is added to FACTERLIB so that
+custom facts may be configured and sent to Foreman. `lib` is added to
+LD_LIBRARY_PATH and `lib/ruby` is added to RUBYLIB, so that binaries in
+`bin` can be executed properly.
+
+Currently, the location of this is hardcoded - the image will look for
+`#{foreman.url}/extension.zip` so place the zip in your Foreman public/
+directory (TODO, add proper routes to the proxy for this).
```
lzap: My initial idea was to accept the following kernel command line options:

    ext=http://any.server.you.like/ext1.zip ext=http://any.server.you.like/ext2.zip ext=http://any.server.you.like/ext3.zip

Since we get the command line verbosely, we can simply allow multiple instances of the same option and read that properly. Or another possibility would be:

    extbase=http://any.server.you.like/extensions ext=ext1.zip ext=ext2.zip

build-livecd
```diff
@@ -6,17 +6,20 @@ VERSION=$(git describe --abbrev=0 --tags)
...
+OVERLAY="/etc/profile.d/discovery.sh
   /etc/rc.d/rc.local
-  /usr/share/fdi/VERSION
-  /usr/share/fdi/RELEASE
-  /usr/share/fdi/facts/discovery-facts.rb
-  /usr/bin/find-missing-libs
-  /usr/bin/find-largest-rpms
+  /etc/systemd/system/discover-host.service
+  /etc/systemd/system/nm-prepare.service
+  /etc/systemd/system/setup-extension.service
```
lzap: Nitpick: I'd prefer `discovery-extensions-setup` but this works too.

root/etc/profile.d/discovery.sh
```diff
@@ -0,0 +1,8 @@
+# /etc/profile.d/discovery.sh - shell extensions for discovery
+
+# Changes to this file should also be reflected in the setup-extension systemd
+# service file, as systemd doesn't source this file
+
+export LD_LIBRARY_PATH=/opt/extension/lib
+export RUBYLIB=/opt/extension/lib/ruby
+export PATH=$PATH:/opt/extension/bin
```
lzap: Do we rather want to append instead of overwriting? What if we install an RPM that actually modifies `LD_LIBRARY_PATH` or `RUBYLIB`. Weird things will hapen then. Do the both in the similar way as the `PATH` please, just to be sure.

root/etc/systemd/system/discover-host.service
```diff
@@ -1,11 +1,11 @@
 [Unit]
 Description=Discover this host to Foreman
 Wants=basic.target
-After=basic.target network.target network-online.target nss-lookup.target
+After=basic.target network.target network-online.target nss-lookup.target setup-extension.service
 
 [Service]
 Type=idle
-Environment="FACTERLIB=/usr/share/fdi/facts"
+Environment="FACTERLIB=/usr/share/fdi/facts:/opt/extension/facts"
```
lzap: Nitpick: Actually we could extract this into a separate file and read that file in both Proxy and this script to avoid copying.

root/etc/systemd/system/setup-extension.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Download and enable zip extension to this image
+Wants=basic.target
+After=basic.target network.target network-online.target nss-lookup.target
+
+[Service]
+Environment=LD_LIBRARY_PATH=/opt/extension/lib
+Environment=RUBYLIB=/opt/extension/lib/ruby
+Environment=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/opt/extension/bin
```
lzap: Idea: Can't we read the `/etc/profile.d/discovery.sh` file here? You would need to split the `export` statement on a separate line and luckily systemd should ignore it. http://www.freedesktop.org/software/systemd/man/systemd.exec.html (`EnvironmentFile`)

root/usr/bin/setup-extension
```diff
@@ -0,0 +1,141 @@
...
+#require 'zip'
+#Zip::File.open("/tmp/ext.zip") do |zipfile|
+#  zipfile.each do |f|
+#    f_path=File.join('/opt/extension', f.name)
+#    FileUtils.mkdir_p(File.dirname(f_path))
+#    zipfile.extract(f, f_path)
+#  end
+#end
+Dir.chdir '/opt/extension'
+system('/bin/unzip','-o','/tmp/ext.zip')
```
lzap: Delete the ZIP after successful download.

root/usr/bin/setup-extension
```diff
@@ -0,0 +1,141 @@
...
+
+# Start automatic script
+if File.exist? '/opt/extension/bin/autostart'
```
lzap: Should be "and it's executable".

root/usr/bin/setup-extension
```diff
@@ -0,0 +1,141 @@
...
+    end
+    if resp.code == 200
+      log_msg "Downloaded extension.zip"
+    else
+      log_err "Problem getting extension.zip: HTTP code #{resp.code}"
+    end
+  end
+ensure
+  f.close()
+end
```
lzap: I'd prefer wget or curl to do this, there might be dragons like HTTP 301. If a tool is not installed, add it to the `*-packages.ks` list, no big deal.

lzap:
Looks great.

README
GregSutcliffe: Agreed, routes are still on the TODO.

build-livecd
GregSutcliffe: That just felt a bit too long - `discovery-setup` ?

root/etc/profile.d/discovery.sh
GregSutcliffe: They're unset by default at the moment. I can change it to be an append, but that won't work in the systemd unit below ($PATH .. etc. doesn't work in systemd)

root/etc/systemd/system/discover-host.service
GregSutcliffe: True, I can do that. On my TODO.

root/etc/systemd/system/setup-extension.service
GregSutcliffe: At the very least we can write our own EnvironmentFile and reference it here. I'll fix it.

root/usr/bin/setup-extension
GregSutcliffe: What difference does it make?

root/usr/bin/setup-extension
GregSutcliffe: That's implicit, system() won't run non-executable stuff (I tested it by accident when I forgot to `chmod +x` on it :P)

root/usr/bin/setup-extension
GregSutcliffe: Fair enough, I was trying to keep it in Ruby where I could. Shelling out is easy enough :)

build-livecd
lzap: > That just felt a bit too long - `discovery-setup` ?

Up to you, or you can just leave it as is.

-- 
Later,
 Lukas #lzap Zapletal

root/etc/profile.d/discovery.sh
lzap: > They're unset by default at the moment. I can change it to be an append, but that won't work in the systemd unit below ($PATH .. etc. doesn't work in systemd)

Please append. You are right with the systemd, I've just checked the
docs. Won't work.

-- 
Later,
 Lukas #lzap Zapletal

root/usr/bin/setup-extension
lzap: > What difference does it make?

More free RAM?

-- 
Later,
 Lukas #lzap Zapletal

root/etc/profile.d/discovery.sh
GregSutcliffe: We've got two options then - either use a hardcoded value in the systemd unit, or use another service unit to write the environment file (this is possible accorind to the systemd.exec man page you linked:

```
The files listed with this directive will be read shortly before the process is executed (more specifically, after all processes from a previous unit state terminated. This means you can generate these files in one unit state, and read it with this option in the next).
```

I quite like this approach actually, we can have `discovery-setup` which downloads the zip and writes the environment file, and then `discover-host` and `discovery-autostart` (new service to run the autostart file) can depend on `discovery-setup` and read the newly written env file. I'll set that up.

root/usr/bin/setup-extension
GregSutcliffe: Fair enough.

GregSutcliffe:
[ ] Extra TODO from irc chat: add a zip template generator

root/etc/profile.d/discovery.sh
ekohl: Maybe you want to prepend instead of append so users can overwrite system commands.

README
```diff
@@ -53,11 +53,39 @@ You can also use the image standalone (without TFTP under Foreman's
...
+
+`bin` is added to PATH, you can place binaries here. If this directory
+contains an executable `autostart` script, it will be run *before* the
+image is registered to Foreman. `facts` is added to FACTERLIB so that
+custom facts may be configured and sent to Foreman. `lib` is added to
+LD_LIBRARY_PATH and `lib/ruby` is added to RUBYLIB, so that binaries in
+`bin` can be executed properly.
+
+Currently, the location of this is hardcoded - the image will look for
+`#{foreman.url}/extension.zip` so place the zip in your Foreman public/
```
lzap: proxy.url

README
lzap: I don't like this hardcoding. Let's just do the ext=HTTP it's straightforward. Users can choose to use proxy/foreman or maybe a different HTTP server on the network.

root/etc/profile.d/discovery.sh
GregSutcliffe: Sounds dangerous, if users really need to do that they can use explicit paths (we can document where the files in up inside the image easily enough)

README
GregSutcliffe: To quote myself: "routes are still on the TODO" and to quote from the description of this PR: "Todo: download location is currently hardcoded - add routes to the proxy/foreman plugins and a PXE option to specify the path"

I'm not sure I can be any clearer - this will be fixed before we merge the PR.

GregSutcliffe:
Ok, the code is all working, with the exception of the routes for downloading it. I see some options:

* Re-use {foreman,proxy}.url and add a new option for a list of names (zips=zip1.zip,zip2.zip)
* Add a new zips.url option with multi-use (zips.url=http://foo.bar/zip1.zip zips.url=https://bar.quux/zip2.url)
* (my favourite) We know we have a TFTP server on the network to serve the Discovery image - we could add the zips there? Then we just use "zips=zip1.zip,zip2.zip..." and get them from TFTP.

All could be implemented with some fallback order, if we wanted, but I'd go for option 3 as it's the minimum setup for the end user (who already has a TFTP server anyway)

build-livecd
GregSutcliffe: Done, renamed to discovery-setup

root/etc/profile.d/discovery.sh
GregSutcliffe: The EnvironmentFile is written by discovery-setup and a new service called discovery-autostart reads that file and runs the script(s). I've also renamed discover-host to discovery-register just to be consistent, as it's a better description of the function.

root/etc/systemd/system/discover-host.service
GregSutcliffe: Done.

root/etc/systemd/system/setup-extension.service
GregSutcliffe: Done.

root/usr/bin/setup-extension
GregSutcliffe: Done.

root/usr/bin/setup-extension
GregSutcliffe: Changed to a bash execution with an `if -x` test for "exists and is executable"

root/usr/bin/setup-extension
GregSutcliffe: Done.

README
GregSutcliffe: Will change once I add the routes.

GregSutcliffe:
New version pushed with the new systemd graph in place. Boot order is:
```
discovery-setup - download zip, extract, write the environment file
  - discovery-autostart - source environment file, run the autostart script from the zip
    - discovery-register (was discover-host) - source the environment file, register host in Foreman
    - foreman-proxy - startup proxy using the environment file
```

note that the proxy and registration are parallel, but both wait until the blocking execution of the autostart has completed.

GregSutcliffe:
First pass at supporting multiple zipfiles added, just need to sort out the generator script and the download routes (as per above). @lzap do you want the zip generator in this repo or the main discovery plugin repo?

lzap:
I like the TFTP idea, easy deployment for us via the Foreman UI. Hopefuly it's reliable. Maybe if you can add a fallback mechanism (zips.url, although I'd prefer just extension).

Graph looks good but I can see we have a bug in there. We should definitely spawn foreman-proxy *before* the discover-host. Can you include the fix in here?

GregSutcliffe:
@lzap graph is fixed, good catch.

TFTP is implemented, just add `fdi.zips=zip1.zip,zip2.zip...` to your APPEND line, and drop them onto the tftp server. The image should query the lease and extract the next-server line, so even if the tftp server isn't on the same host as the dhcp-server, it should still work. I've tested it as thoroughly as I can, but please try it yourself :)

Example zip structure added to `example_zip` - you can zip it up via `cd example_zip && zip -r zip1.zip .`

README updated but could probably be worded better - suggestions welcome.

README
```diff
@@ -53,11 +53,57 @@ You can also use the image standalone (without TFTP under Foreman's
...
+appended to. If you need to specify the path to something explicitly in
+your scripts, the zip contents are extracted to `/opt/extension` on the
+image.
+
+An example structure is provided in `example_zip` in this repo. You can zip
+up your structure with `zip -r my_extension.zip .`
+
+You can create multiple zip files, but be aware they will be extracted to
+the same place on the discovery image, so files in later zips will overwrite
+earlier ones if they have the same filename.
```
lzap: Hmmm including the `autostart` script? Will need to check.

example_zip/autostart.d/01_zip.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+echo "Discovery Extension zipfile loaded" > /tmp/discovery-extension.log
+
+# This example shows the RUBYLIB extension working
+/usr/bin/ruby -rdiscovery_extension -e'puts Discovery::Extension.confirm'
```
lzap: Oh I see its a directory now. Good.

root/usr/bin/get-zip-server_config
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Script to extract the NEXT_SERVER value from the incoming DHCP lease
+# and write it to a place where systemd/profile.d can source it
+
+# NetworkManager automatically strips ".sh" and appends "_config" on the source
+# file. This is only run when the interface is going "up" so we don't interfere
+# with the next-server settings on "down"
```
lzap: Interesting, why you didn't put this into the `.sh` script directly? What is the advantage here?

root/etc/profile.d/discovery.sh
```diff
@@ -0,0 +1,28 @@
+# /etc/profile.d/discovery.sh - shell extensions for discovery
+
+# Parse the systemd EnvironmentFile(s) and export the values for use
+# in normal shells
```
lzap: Nice.

root/etc/systemd/system/discovery-setup.path
```diff
@@ -0,0 +1,11 @@
...
+# Network-Manager-wait-online doesn't seem to wait *quite* long enough
+# for the custom post-up command to write the next-server environment
+# file. So we use a path target to wait for it to be written.
+Description=Waiting for ZIP_SERVER environment file to be created
+
+[Path]
+PathExists=/etc/default/discovery-zip-server
+
+[Install]
+WantedBy=multi-user.target
```
lzap: This is nice solution!

root/usr/bin/discovery-autostart
```diff
@@ -0,0 +1,36 @@
...
+
+# Since we allow multiple zipfile extensions, we require users to put
+# autostart scripts in autostart.d, which will then be executed in
+# POSIX order, in blocking mode so that they can complete before the
+# host is registered. Users are responsible for backgrounding long tasks
+# in their own startup scripts.
+
+if test -d /opt/extension/autostart.d; then
+  for script in /opt/extension/autostart.d/*; do
+    test -x "$script" && $script
```
lzap: Output should go to journald I suppose. Will check that during testing.

lzap:
Looks great, testing now.

README
GregSutcliffe: Yes, the whole think is clobbered. Last unpacked zip will win.

example_zip/autostart.d/01_zip.sh
GregSutcliffe: Indeed, thought of that :)

root/usr/bin/get-zip-server_config
GregSutcliffe: It's how NetworkManager expects things to be structured. `.sh` is sourced on both `up` and `down` events, where as `_config` is executed only on `up`.

root/etc/systemd/system/discovery-setup.path
GregSutcliffe: Yeah, wait-online didn;t wait *quite* long enough. I was about to add a sleep when I saw that you could do this instead :)

root/usr/bin/discovery-autostart
GregSutcliffe: Hopefully, if it doesn't we can explicitly add it.

build-livecd
```diff
@@ -37,11 +46,15 @@ done
...
+  /usr/bin/discovery-autostart \
   /usr/bin/discovery-debug \
+  /usr/bin/discovery-register \
+  /usr/bin/discovery-setup \
+  /usr/bin/find-largest-rpms \
+  /usr/bin/find-missing-libs \
+  /usr/bin/get-zip-server_config
   /usr/bin/nm-prepare
 %end
 KS
```
lzap: This is strange, I've upgraded to Fedora 21 and ksflatten now fails requiring me to specify `-c` option. It looks like this option was there for years. Can you please add it? So instead of

    ksflatten "$1" -o $KS_TARGET

do

    ksflatten -c "$1" -o $KS_TARGET

and check it works for you as well. Thanks.

lzap:
It works just fine! Ready to merge. Please squash into the final one. Good work.

root/usr/bin/discovery-autostart
lzap: Yes its in the systemd docs.

root/usr/bin/get-zip-server_config
lzap: Hmmm but "up" or "down" are passed as the first parameter to the `.sh` script. I use it on my laptop locally. Anyway, this works for me.

build-livecd
GregSutcliffe: Hah, I had to add that myself, and then took it out again when I sent the PR :P

GregSutcliffe:
@lzap "-c" added and squashed. Can you add the issue number on merge?

lzap:
Merged for Discovery 2.0 (image 2.1). Thanks. e74dc86

PR #7, opened 2015-03-18
Fixes #9810 - IPMI facter extensions multichannel support

lzap:
This patch now fixes ipmi_facts and also add ipmi_N_facts (where N is channel
number) for those who would like to get information from multiple channels for
any reason.

Example run of facter with this patch on IBM server with IPMI on channel 1:

```
[root@ibm-x3655-01 ~]# FACTERLIB=/root facter | grep ipmi
ipmi_1_gateway => 10.16.71.254
ipmi_1_ipaddress => 10.16.68.91
ipmi_1_ipaddress_source => DHCP Address
ipmi_1_macaddress => 00:14:5e:5b:8c:8e
ipmi_1_subnet_mask => 255.255.248.0
ipmi_enabled => true
ipmi_gateway => 10.16.71.254
ipmi_ipaddress => 10.16.68.91
ipmi_ipaddress_source => DHCP Address
ipmi_macaddress => 00:14:5e:5b:8c:8e
ipmi_subnet_mask => 255.255.248.0
```

The same on HP box with IPMI on channel 2:

```
[root@hp-dl165g6-01 ~]# FACTERLIB=/root facter | grep ipmi
ipmi_2_gateway => 10.16.71.254
ipmi_2_ipaddress => 10.16.66.131
ipmi_2_ipaddress_source => DHCP Address
ipmi_2_macaddress => 00:25:b3:e6:5d:98
ipmi_2_subnet_mask => 255.255.248.0
ipmi_enabled => true
ipmi_gateway => 10.16.71.254
ipmi_ipaddress => 10.16.66.131
ipmi_ipaddress_source => DHCP Address
ipmi_macaddress => 00:25:b3:e6:5d:98
ipmi_subnet_mask => 255.255.248.0
```

@mhulan can you take a look?

PR #9, opened 2015-04-07
Fix missing nm-prepare executable bits

beddari:
Looks like we're currently missing a line escape:

```
[root@fdi bin]# ls -l /usr/bin/nm-prepare
-rw-r--r-- 1 root root 719 Jan 14 12:00 /usr/bin/nm-prepare
```

```sh
Apr 07 16:49:13 fdi systemd[1161]: Failed at step EXEC spawning /usr/bin/nm-prepare: Permission denied
Apr 07 16:49:13 fdi systemd[1]: nm-prepare.service: main process exited, code=exited, status=203/EXEC
Apr 07 16:49:13 fdi systemd[1]: Unit nm-prepare.service entered failed state.
```

### DIFF
--- a/build-livecd
+++ b/build-livecd
@@ -54,7 +54,7 @@ chmod +x \
   /usr/bin/discovery-setup \
   /usr/bin/find-largest-rpms \
   /usr/bin/find-missing-libs \
-  /usr/bin/get-zip-server_config
+  /usr/bin/get-zip-server_config \
   /usr/bin/nm-prepare
 %end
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
Good catch! :+1:

lzap:
Thanks I would like to merge https://github.com/theforeman/foreman-discovery-image/pull/7 and then release 2.2 version of the image. Merged as 3d4e312

PR #12, opened 2015-05-20
Fixes #10562 - nm-prepare sends output to stdout

lzap:


lzap:
Reproduce: Boot the current image and to `systemctl status nm-prepare`

lzap:
This was caused by journald/systemd. I was sending data through `systemd-cat` and this was not warning in early boot. Apparently systemd by default acquires STDOUT/STDERR of the processes/services, so there was no point of doing that at all. This is cleaner.

On top of that, I realized that reboots are slow, some process is stuck and I don't know which one. This limits the delay to 5 seconds before the process is killed.

root/etc/systemd/system/nm-prepare.service
```diff
@@ -5,6 +5,7 @@ Before=NetworkManager.target network-online.target
 
 [Service]
 Type=oneshot
+RemainAfterExit=yes
```
lzap: This is cosmetic thing, we have the same for all the other `oneshot` services. AFAIK this only affects how systemd reports this service.

stbenjam:
@lzap Looks good, sorry for the delay.  `systemctl status nm-prepare` shows `active (exited)` and the correct log entries now.

Would you like me to merge this? I never use the Foreman procedure for merging - is it just a direct cherry-pick + push to develop or are there docs somewhere for it?

lzap:
You can merge via github here I think, or do ff if you like. No strict rules.

lzap:
Thanks and safe flight!

stbenjam:
Thanks! :wave:

lzap:
Thanks, since I will be backporting this into 2.1 stable branch, --no-ff might be better. Next time! :)

stbenjam:
Ah will do, that's why I asked :-) You should be able to cherry-pick the (non-merge) commit to your 2.1 branch cleanly.

PR #13, opened 2015-05-22
Updated README build instructions

stbenjam:
Mostly just wanted to add the package requirements (pykickstart and livecd-tools) so a new dev doesn't have to wonder.  

But also moved this to be truly Markdown. I'm happy to revert that if you'd prefer plain text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-
+```
   _____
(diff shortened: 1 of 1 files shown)


lzap:
Merged as 44d30d92db57bba579e5678b22f74a0d6fc25639, thank you!

PR #14, opened 2015-06-12
Fixes #10650 - fixed race condition in discovery-setup

lzap:
See the ticket for more info.

### DIFF
--- a/root/usr/bin/discovery-setup
+++ b/root/usr/bin/discovery-setup
@@ -54,10 +54,10 @@ unless zips.nil?
       else
         log_msg "TFTP error downloading #{zip}: #{tftp_result}"
       end
-
     rescue
       # We always want to write /etc/default/discovery so don't bomb out here
       log_err "Problem downloading zipfile #{zip}"
+      exit 1
     ensure
       # Cleanup
       File.unlink('/tmp/ext.zip') if File.exists?('/tmp/ext.zip')
@@ -73,3 +73,5 @@ File.open('/etc/default/discovery','w') do |f|
   f.write(env_append('LD_LIBRARY_PATH','/opt/extension/lib'))
   f.write(env_append('RUBYLIB','/opt/extension/lib/ruby'))
 end
+
+exit 0
--- a/root/etc/systemd/system/discovery-setup.service
+++ b/root/etc/systemd/system/discovery-setup.service
@@ -1,10 +1,12 @@
 [Unit]
 Description=Download zip extensions to this image
 Wants=basic.target
-After=basic.target network.target network-online.target nss-lookup.target
+After=basic.target network.target network-online.target nss-lookup.target discovery-setup.path
 
 [Service]
 Type=oneshot
+Restart=on-failure
+RestartSec=5
 EnvironmentFile=-/etc/default/discovery-zip-server
 RemainAfterExit=yes
 ExecStart=/usr/bin/discovery-setup


lzap:
The build will appear in 40 minutes in

http://downloads.theforeman.org/discovery/lzap/scratch/

Can you test please? Thank you!

japz:
Yes, I will be able to test it early next week.

lzap:
Build done. Ignore the version of the image, I forgot to push a tag.

lzap:
@stbenjam mind doing review next week? Thanks!

aux/vagrant-build/Vagrantfile
```diff
@@ -6,6 +6,8 @@ boxes = [
   {:name => 'wheezy',   :libvirt => 'fm-debian7',    :image_name => /Debian.*7/,      :os_user => 'debian'},
   {:name => 'f19',      :libvirt => 'fm-fedora19',   :image_name => /Fedora.*19/, :pty => true},
   {:name => 'f20',      :libvirt => 'fm-fedora20',   :image_name => /Fedora.*20.*PVHVM/, :pty => true},
+  {:name => 'f21',      :libvirt => 'fm-fedora21',   :image_name => /Fedora.*21.*PVHVM/, :pty => true},
+  {:name => 'f22',      :libvirt => 'fm-fedora22',   :image_name => /Fedora.*22.*PVHVM/, :pty => true},
```
stbenjam: Should this not be a separate commit? Doesn't seem related.

japz:
Doesn't fix it unfortunately (still sometimes fails).

I think discovery-setup should exit with a non-zero exit code if it fails to fetch any of the zip-files and then we could have systemd retry.

stbenjam:
I do think discovery-setup should fail instead of always returning successful., maybe with the addition of a retry mechanism.

@japz What's going on that the download routinely fails though?

japz:
I think it has to do with the fact that we have STP (or some variant of that) enabled on our switches. The DHCP response may be delayed because the switch port is not yet forwarding and discovery-setup runs before its environment exists.

An alternative (and arguably simpler) approach would be to get the server that should be used for downloading the zips from /proc/cmdline.

aux/vagrant-build/Vagrantfile
lzap: Yeah right, actually our Jenkins/Raskspace ceased working from the last week (they deleted old Fedora versions). I will do separate one.

lzap:
Ok the image is rebuilding with the restart trick. At least this helps from incorrect behavior. Please re-test.

http://ci.theforeman.org/view/Packaging/job/packaging_discovery_image/29/

root/usr/bin/discovery-setup
```diff
@@ -54,10 +54,10 @@ unless zips.nil?
       else
         log_msg "TFTP error downloading #{zip}: #{tftp_result}"
       end
-
     rescue
       # We always want to write /etc/default/discovery so don't bomb out here
       log_err "Problem downloading zipfile #{zip}"
+      exit 1
```
japz: Save the exit code for later (to exit with after writing the env-files), or remove this comment.

japz:
Tested. Discovery now fails because /etc/default/discovery does not exist.

lzap:
Mind pastebining the journald log (interested in ordering).

lzap:
Just for the record, there is a different patch that workarounds this bug: https://github.com/theforeman/foreman-discovery-image/pull/15

I will attempt to fix the root cause of this later on here.

lzap:
Ok for the record, the current patch does not work because of broken dep:

```
Jun 16 09:35:06 fdi systemd[1]: discovery-setup.service has Restart setting other than no, which isn't allowed for Type=oneshot services. Refusing.
Jun 16 09:35:06 fdi systemd[1]: Cannot add dependency job for unit discovery-setup.service, ignoring: Unit discovery-setup.service failed to load: Invalid argument. See system logs and 'systemctl status discovery-setup.service' for details.
```

lzap:
Ok I think we actually need `PathChanged=`: may be used to watch a file or directory and activate the configured unit whenever it changes. It is not activated on every write to the watched file but it is activated if the file which was open for writing gets closed.

http://www.freedesktop.org/software/systemd/man/systemd.path.html

I think we read the file when it gets created (and is empty).

lzap:
There is a different patch emerging that fixes the fact we forgot to do After=wait-online service and it might actually fix this bug.

lzap:
Leaving this opened since I think we need still to fix `PathChanged`.

lzap:
https://github.com/theforeman/foreman-discovery-image/pull/18 looks like a better solution to this problem. Closing this one for now.

PR #15, opened 2015-06-16
Fixes #10831 - introduced fdi.zipserver command line option

japz:
Adds support for fdi.zipserver= to force the server the discovery image should download the extensions from.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,10 @@ paths are relative to the TFTP root. So if you have two zips at
 $TFTP/zip1.zip and $TFTP/boot/zip2.zip, you would use
 `fdi.zips=zip1.zip,boot/zip2.zip`.
 
+You can force the server to download the zip files by appending
+`fdi.zipserver=<tftp-address>`, where `<tftp-address>` is the IP address of
+your TFTP server hosting the extensions.
+
 Planned features
 ----------------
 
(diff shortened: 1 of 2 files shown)


lzap:
Ok created http://projects.theforeman.org/issues/10831. Can you amend the commit message to:

    Fixes #10831 - introduced fdi.zipserver command line option

I will trigger the build.

lzap:
Building ... http://downloads.theforeman.org/discovery/lzap/japz/

lzap:
Success!

stbenjam:
Is there a reason to limit it to TFTP? It might be nice to support any arbitrary URI with cURL.  Could do it later, but just a thought.

japz:
Works!

@stbenjam I agree.

japz:
Changed the commit message as requested.

lzap:
@stbenjam the idea is to leverage smart proxy for this and we are currently running TFTP service file hosting there. In future, the plan is to create some kind of UI for image "extensions", Foreman can deploy those extensions via smart-proxy.

@japz I think this is a reasonable solution and it is documented now, thanks.

lzap:
Merged as 19ebe946b97f08e1eca7c82ba29f6bc836dd1bd9, thank you!

PR #17, opened 2015-07-16
Fixes #11134 - introduced fdi.initnet option

lzap:


lzap:
http://ci.theforeman.org/view/Packaging/job/packaging_discovery_image/34/

20-packages.ks
```diff
@@ -11,6 +11,7 @@ yum
 vim-minimal
 acpid
 tftp
+lldpad
```
lzap: We also add this small package for Intel LLDPAD

lzap:
http://downloads.theforeman.org/discovery/lzap/initnet/

balexx:
nm-prepare works correctly and generates the expected ifcfg-eno1 and ifcfg-eno2 files - if you run it manually. It doesn't seem to run on boot - both ifcfg files are missing from /etc/sysconfig/network-scripts prior to manually executing the script.

lzap:
Let me test this, it is supposed to run.

lzap:
Ok we found a bug - nm-prepare sometimes starts *before* network cards are initialized via `sys-subsystem-net-devices-eth0.device`. We need to prevent the ordering here.

lzap:
Okay @balexx here is another build. I have rewritten the way we initialize network and actually since multiple network cards is causing lot of troubles, I am making fdi.initnet=bootif the default option. New image is building, download it from the same place: http://ci.theforeman.org/view/Packaging/job/packaging_discovery_image/36/

root/etc/profile.d/discovery.sh
```diff
@@ -4,25 +4,31 @@
 # in normal shells
 
 filename="/etc/default/discovery"
-while read -r line
-do
+if test -f $filename
```
lzap: This was just reformatted and if the `$filename` is not present (extensions are not enabled), then this does not throw error during login (profile load). That was just annoying.

lzap:
Reviewers @stbenjam @orrabin: I got rid of rh-plugin in NetworkManager and we now configure NM not to start interfaces automatically. Instead, nm-prepare is called via systemd.service *before* NetworkManager starts with `primary` param which drops primary interface configuration.

Also if `fdi.netinit=all` is provided, then via udev configurations for all remaining network cards is also generated. This happens usually *after* NM was started, so it automatically picks connections as they come up and brings that.

lzap:
This is how it looks like (default behavior):

```
[root@fdi ~]# nmcli con
NAME                         UUID                                  TYPE            DEVICE 
secondary-52:54:00:19:7d:d1  74135b24-2c81-11e5-bc44-525400130513  802-3-ethernet  --     
secondary-52:54:00:46:c0:fb  73e159b2-2c81-11e5-b5e2-525400130513  802-3-ethernet  --     
primary                      7305b150-2c81-11e5-8a1a-525400130513  802-3-ethernet  eth0   
secondary-52:54:00:15:f8:b8  7413d8f6-2c81-11e5-aa20-525400130513  802-3-ethernet  --     

[root@fdi ~]# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 52:54:00:13:05:13 brd ff:ff:ff:ff:ff:ff
    inet 192.168.122.68/24 brd 192.168.122.255 scope global dynamic eth0
       valid_lft 3587sec preferred_lft 3587sec
    inet6 fe80::5054:ff:fe13:513/64 scope link 
       valid_lft forever preferred_lft forever
3: ens4: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 52:54:00:46:c0:fb brd ff:ff:ff:ff:ff:ff
4: ens5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 52:54:00:15:f8:b8 brd ff:ff:ff:ff:ff:ff
5: ens6: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether 52:54:00:19:7d:d1 brd ff:ff:ff:ff:ff:ff
```

balexx:
updated image works perfectly!

lzap:
> updated image works perfectly!

Have you tested `all` option as well? I'd be interested if it
initializes ifaces as expected.

-- 
Later,
 Lukas #lzap Zapletal

balexx:
Tested and works as expected - both interfaces get an IP

lzap:
To test this build and run fdi image with standard options, one interface should be initialized via DHCP. Then run with `all` option and all of them should be initialized via DHCP. I'd like to have this in the new fdi release. Thanks!

lzap:
@balexx thanks for testing, the above comment was not for you but for devs with push access like @stbenjam or @orrabin ;-)

stbenjam:
Tested and works well, merged as d4cb4ea7f4f84396542ebd1e9c905e032e11cedd

PR #18, opened 2015-07-21
Wait for usable network connectivity before running discovery-*

schiermi:
In systems with many (network & block) devices or in networks with slow DHCP servers the discovery-* services start without a proper resolv.conf & network connectivity.

This leads to:
-failing TFTP downloads of extension zips (server not reachable or name resolution error in case address supplied via fdi.zipserver is a hostname).
-discovery-registers is not able to register against katello/foreman because /etc/resolv.conf is not usable when the ruby script starts / tries name resolution for the first time (resolv.conf ist not re-read by Ruby / the script)

The directive "After=…network-online.target…" in the current discovery-*service definitions is not sufficient. This target (at least in Centos 7) is reached as soon NetworkManager is started.

This diff adds another service (and correspondign Requires to discovery-*.services) which returns as soon NetworkManager thinks the system is connected to a network.


lzap:
Hello, thanks for your contribution!

How does this differ from a service that ships with RHEL7/CentOS7 and which is already enabled on our image?

```
cat /usr/lib/systemd/system/NetworkManager-wait-online.service
[Unit]
Description=Network Manager Wait Online
Requisite=NetworkManager.service
After=NetworkManager.service
Wants=network.target
Before=network.target network-online.target

[Service]
Type=oneshot
ExecStart=/usr/bin/nm-online -s -q --timeout=30

[Install]
WantedBy=multi-user.target
```

Also discovery-register can start without network, it will just fail with the initial request, but it re-sends every minute.

Since smart-proxy binds to all interfaces it should be also fine with starting before network was initialized.

The only issue is the TFTP, we have a different PR that tries to solve it, but does not yet work: https://github.com/theforeman/foreman-discovery-image/pull/14

lzap:
I mean your patch is still valid, but can you delete unnecessary `network.target` and `network-online.target` and replace that with `NetworkManager-wait-online.service` instead?

More reading on this topic is at: http://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

lzap:
The bug might be in the fact that we have the service enabled, but we need to order our services after that online service.

schiermi:
/usr/lib/systemd/system/NetworkManager-wait-online.service calls nm-online with parameter -s:

man 1 nm-online:
```
       -s, --wait-for-startup
              Wait for NetworkManager startup to complete, rather than waiting for network connectivity specifically.  Startup
              is  considered complete once NetworkManager has activated (or attempted to activate) every auto-activate connec‐
              tion which is available given the current network state. (This is generally only  useful  at  boot  time;  after
              startup has completed, nm-online -s will just return immediately, regardless of the current network state.)
```
"Wait for NetworkManager startup to complete, rather than waiting for network connectivity specifically." is not sufficient. nm-connected.service calls nm-online without -s (and a higher timeout) which is much more meaningful before running discovery-*.

discovery-register tries every 30 seconds, and fails because of "getaddrinfo: Name or service not known" when trying to resolv our Katello/Foreman hostname. The retry works when using IP addresses. When using DNS the first lookup fails (because /etc/resolv.conf is missing / incomplete) and subsequent requests fail also because ruby does not re-initialize (res_init()) the resolver, which still uses the broken resolv.conf information (https://sourceware.org/bugzilla/show_bug.cgi?id=984).

nm-connected.service is just a drop-in to ensure all discovery-* services have proper network connectivity available the time they are started.

My PR also solves the TFTP issue without depending on file "semaphores".

lzap:
Ok that looks good. Can you please squash into one commit with msg:

    Fixes #10650 - wait for nm IPv4 DHCP initialization

Also couple of other comments will follow.

22-discovery.ks
```diff
@@ -15,6 +15,9 @@ systemctl enable NetworkManager-wait-online.service
 echo " * enabling nm-prepare service"
 systemctl enable nm-prepare.service
 
+echo " * enabling nm-connected service"
+systemctl enable nm-connected.service
+
```
lzap: Also remove `systemctl enable NetworkManager-wait-online.service` block please as this replaces it completely.

root/etc/systemd/system/discovery-autostart.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=Start downloaded extension scripts
+Requires=nm-connected.service
```
lzap: Why this? We do not need to activate `nm-connected` when autostart (and others) are activated. I don't follow why we need this:

```
       Requires=
           Configures requirement dependencies on other units. If this unit gets activated, the units listed here will be
           activated as well. If one of the other units gets deactivated or its activation fails, this unit will be
           deactivated. This option may be specified more than once or multiple space-separated units may be specified in
           one option in which case requirement dependencies for all listed names will be created. Note that requirement
           dependencies do not influence the order in which services are started or stopped. This has to be configured
           independently with the After= or Before= options. If a unit foo.service requires a unit bar.service as
           configured with Requires= and no ordering is configured with After= or Before=, then both units will be started
           simultaneously and without any delay between them if foo.service is activated. Often it is a better choice to
           use Wants= instead of Requires= in order to achieve a system that is more robust when dealing with failing
           services.

           Note that dependencies of this type may also be configured outside of the unit configuration file by adding a
           symlink to a .requires/ directory accompanying the unit file. For details see above.
```

root/etc/systemd/system/nm-connected.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Network Manager Wait Connected
```
lzap: Maybe something like this could be better: `Wait for NetworkManager DHCP activation` because without `-s` this is really awaiting DHCP activation rather than awaiting connection.

root/etc/systemd/system/nm-connected.service
```diff
@@ -0,0 +1,11 @@
...
+Requisite=NetworkManager.service
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/nm-online -q --timeout=120
+RemainAfterExit=true
+
```
lzap: Trail line.

lzap:
Unfortunately we cannot remove the dance around TFTP downloading. We need to put `/etc/default/discovery` into both `foreman-proxy` and `discovery-register` and activated nm-online does not necessary mean that all dispatch scripts has been successfuly executed. Therefore we need to do it this way.

But there is one bug still, can you replace `PathExists` with `PathChanged` in `discovery-setup.service` please? This ensures the unit gets activated after the file is *closed* and not *opened*.

If you find a different way to make sure we have the environment variables ready for both services above, send your proposal as a different commit please (with the same issue number).

root/etc/systemd/system/discovery-autostart.service
schiermi: Requires= is critical to ensure all the discovery-….service are not started before nm-connected.service finished **successfully**. Without Requires= (only with After=) the services would start regardless of nm-online's return code.

22-discovery.ks
schiermi: Without NetworkManager-wait-online.service we effectively lose the synchronization point in network-online.target. Or do I miss something?

root/etc/systemd/system/nm-connected.service
schiermi: d'accord.

root/etc/systemd/system/nm-connected.service
schiermi: d'accord.

22-discovery.ks
lzap: Hmmm I thought that since `nm-connected` have `Before=network-online.target` we don't need that.

root/etc/systemd/system/discovery-autostart.service
lzap: I see, ok. Well if `nm-online` fails to start, we are in trouble anyway, but good point.

lzap:
Ok it works fine, if you can do the `PathChanged` change, nit picks and rebase, I am happy to merge for the next release.

22-discovery.ks
schiermi: You are right.
/usr/lib/systemd/system/network-online.target.wants/NetworkManager-wait-online.service comes from the NetworkManager package and is NOT provided by NetworkManager-wait-online.service [Install]-section. I remove `systemctl enable nm-connected.service` and give it a try.

lzap:
Or we can leave `PathChanged` out of this patch perhaps, but I was under impression this can also cause race condition.

schiermi:
Even with PathChanged a race condition exists: discovery-setup.service is pulled by multi-user.target (after all units up to nm-connected.service triggered). The unit does not complain about an missing EnvironmentFile (there is a - in `EnvironmentFile=-/etc/default/discovery-zip-server`) which may have not been created at this point in time. In case fdi.zipserver is not defined, we do not get the zips inside /usr/bin/discovery-setup. Let me think a moment about that…

lzap:
Yeah `/etc/default/discovery-zip-server` file is treated basically as a semaphore. We could simply ship it as it is unchanged by the script. But we must be sure that discover-setup *finishes* before discovery-register and foreman-proxy are started. I don't know how to achieve this with systemd.

schiermi:
I suggest to move the kernel command line parsing out of discovery-setup and put it into get-zip-server_config.
get-zip-server_config should check if the DHCP response is for the interface the system was booted from (BOOTIF kernel command line). Afterwards this script should create the /etc/default/discovery-zip-server either from the kernel parameter or from DHCP's next-server. /etc/default/discovery-zip-server is monitored by systemd to start the subsequent discovery-* units.

I'm going prepare a proposed patchset within one day.

schiermi:
This way round PathExists does not need to be modifed.

lzap:
> I'm going prepare a proposed patchset within one day.

Ok, also I was wondering why we use dhclient script and not
`/etc/NetworkManager/dispatcher.d` standard NM script. This way we could
rid of the weird naming `_config`.

-- 
Later,
 Lukas #lzap Zapletal

lzap:
> This way round PathExists does not need to be modifed.

I'd still rather see `PathChanged` tho:

```
       PathExists=, PathExistsGlob=, PathChanged=, PathModified=, DirectoryNotEmpty=
           Defines paths to monitor for certain changes: PathExists= may be used to watch the mere existence of a file or
           directory. If the file specified exists, the configured unit is activated.  PathExistsGlob= works similar, but
           checks for the existence of at least one file matching the globbing pattern specified.  PathChanged= may be used
           to watch a file or directory and activate the configured unit whenever it changes. It is not activated on every
           write to the watched file but it is activated if the file which was open for writing gets closed.  PathModified=
           is similar, but additionally it is activated also on simple writes to the watched file.  DirectoryNotEmpty= may
           be used to watch a directory and activate the configured unit whenever it contains at least one file.

```
-- 
Later,
 Lukas #lzap Zapletal

lzap:
I'd also removed `-` characters in Env statements, we should fail out early. If it does not exist, that's a problem. Maybe to statically ship this file and use just some empty file semaphore instead?

schiermi:
Maybe the dhclient script was intended with some compatibility in mind? On the other hand get-zip-server_config is not "pure-dhclient-ready" because it uses DHCP4_… variables instead of dhclients new_dhcp4_…. Im going to rearrange the functionality to be placed into /etc/NetworkManager/dispatcher.d.

schiermi:
I've reorganized the startup of the discovery-* scripts. Their units are not longer "WantedBy" multi-user.target. Instead the chain starting at discovery-setup.service is triggered by "PathExists=/etc/default/discovery-zip-server" in discovery-setup.path. discovery-setup.service wants discovery-autostart.service which wants discovery-register.service. Other dependencies are dropped (including my nm-connected.service.

/etc/default/discovery-zip-server is created by /etc/NetworkManager/dispatcher.d/15-get-zip-server; a "pure" NM dispatcher.d script. This script only processes "up" events for the "BOOTIF" interface.
The kernel cmdline parameter fdi.zipserver takes precedence over the DHCP supplied next server. Using an temporary file only a non-empty file is written (to match systemd PathExists).

nm-prepare was also prone to race conditions. I replaced the systemd unit with a more appropriate udev rule and modified the script.

lzap:
Re: dhclient compatibility - no backward compatibility needed, our image only supports CentOS7/RHEL7+.

Taking a look, btw can you rebase it looks like there is a small conflict. Thanks.

root/usr/bin/nm-prepare
```diff
@@ -4,21 +4,23 @@ echo "Kernel command line: $(cat /proc/cmdline)"
...
+else
+  echo "Configuring interface ${dev}"
+  peerdns=no
+  defroute=no
+fi
+script=/etc/sysconfig/network-scripts/ifcfg-$dev
+cat > $script <<EONS
 DEVICE=$dev
 TYPE=Ethernet
 BOOTPROTO=dhcp
```
lzap: Damn, this whole file has been completely rewritten in d4cb4ea7f4f84396542ebd1e9c905e032e11cedd and it also triggers via udev now. You can basically drop your changes as it works the same way (accepts a parameter = device name). It now does not create rh-cfg plugin configurations but directly NetworkManager ini-style profiles instead. I also use `systemd-cat` utility to redirect output to journal.

root/etc/udev/rules.d/98-nmprepare.rules
```diff
@@ -0,0 +1 @@
+SUBSYSTEM=="net", KERNEL!="lo", RUN+="/usr/bin/nm-prepare $name"
```
lzap: This can go as well as the rule is already there (in d4cb4ea7f4f84396542ebd1e9c905e032e11cedd).

root/etc/systemd/system/discovery-setup.service
```diff
@@ -1,13 +1,10 @@
...
 
 [Service]
 Type=oneshot
-EnvironmentFile=-/etc/default/discovery-zip-server
+EnvironmentFile=/etc/default/discovery-zip-server
 RemainAfterExit=yes
 ExecStart=/usr/bin/discovery-setup
-
-[Install]
-WantedBy=multi-user.target
```
lzap: To be honest, I was never fan of `discovery-setup`. It's so misleading, I'd prefer `discovery-fetch-extensions` or something... ;-)

root/etc/systemd/system/discovery-autostart.service
```diff
@@ -1,14 +1,10 @@
...
 
 [Service]
 Type=oneshot
-EnvironmentFile=-/etc/default/discovery
+EnvironmentFile=/etc/default/discovery
 RemainAfterExit=yes
 ExecStart=/usr/bin/discovery-autostart
-
-[Install]
-WantedBy=multi-user.target
```
lzap: Hmm maybe `discovery-start-extensions` too. Anyway, it looks good.

root/etc/NetworkManager/dispatcher.d/15-get-zip-server
```diff
@@ -0,0 +1,62 @@
...
+      # retrieve zipserver from the kernel cmdline
+      zipserver="${value#*=}"
+    ;;
+    "BOOTIF")
+      # normalization of MAC addresses (lower case, separated by :)
+      # strip out leading arp type (01-)
+      bootif="$(tr A-F- a-f: <<< "${value#*-}")"
+    ;;
+  esac
+done
```
lzap: Nitpick: Can you make this a generic function putting it in `/usr/share/fdi` perhaps? All `fdi.something` values setting as shell variables.  This could be reused in nm-prepare.

lzap: Maybe a prefix like `KCL_` and all shell variables uppercase to prevent injection ;)

lzap:
Ok this looks good. Please note the new `fdi.initnet` option after you rebase. We now initialize only primary interface *by default*. Test this with `fdi.initnet=all` option as well to see if it prevents from ordering bug.

lzap:
During rebase, please reword the commit message to `Fixes #11241 - ...`

http://projects.theforeman.org/issues/11241

lzap:
Oh one note, `nm-prepare` is being called from udev, but to configure the primary interface it is still being called as systemd service. A special parameter `primary` is provided in that case and MAC address is taken from the kernel command line (appended there by PXELinux).

schiermi:
Sounds good.
I will rename…
-discovery-setup to discovery-fetch-extensions
-discovery-autostart to discovery-start-extensions

Introduce /usr/share/fdi/commonfunc.sh with functions exportKCL and normalizeHwAddr.
I'm going to replace all occurrences of /proc/cmdline parsing with commonfunc.sh / exportKCL.

schiermi:
KCL_BOOTIF is the variable based on the kernel cmd line (including ARP type 01).
BOOTMAC is derived from this with the ARP type stripped.

lzap:
> KCL_BOOTIF is the variable based on the kernel cmd line (including ARP type 01).
> BOOTMAC is derived from this with the ARP type stripped.

Sounds great.

-- 
Later,
 Lukas #lzap Zapletal

schiermi:
Image based in Centos 7 tested successfully with and without fdi.initnet=all.

root/etc/rc.d/rc.local
```diff
@@ -1,12 +1,14 @@
...
-  echo "root:$ROOTPW" | chpasswd >/dev/null
+if [[ ! -z "$KCL_FDI_ROOTPW" ]]; then
+  echo "root:$KCL_FDI_ROOTPW" | chpasswd >/dev/null
 fi
 
 # enable ssh
-if [[ "$(cat /proc/cmdline | grep -io 'fdi.ssh=\S*')" == "fdi.ssh=1" ]]; then
+if [ "$KCL_FDI_SSH" == "1" ]]; then
   systemctl start sshd
 fi
```
lzap: This does not work for me, sshd is not started.

lzap: It looks like commonfunc.sh does work, but rc.local is not enabled. This needs to be enabled explicitly on systemd systems AFAIK.

root/usr/bin/discovery-fetch-extensions
```diff
@@ -70,7 +70,7 @@ end
 
 # Write out environment files for subsequent paths and systemd units to use
 File.open('/etc/default/discovery','w') do |f|
-  f.write("# EnvironmentFile generated by discovery-setup.service\n")
+  f.write("# EnvironmentFile generated by discovery-fetch-extensions.service\n")
   f.write(env_append('PATH','/opt/extension/bin'))
   f.write(env_append('FACTERLIB','/usr/share/fdi/facts:/opt/extension/facts'))
   f.write(env_append('LD_LIBRARY_PATH','/opt/extension/lib'))
```
lzap: Do we still need to create this file on-the-fly? I mean, can't we hardcode it into the image since we don't change it and there is no path hooked to it?

root/usr/share/fdi/commonfunc.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/echo This file is meant to sourced not executed:
```
lzap: Nice trick! :-)

root/etc/rc.d/rc.local
lzap: Usually make sure it's executable and then `systemctl enable rc-local.service`.

lzap: Oh wait its a syntax error actually: `Jul 31 10:51:55 fdi rc.local[1107]: /etc/rc.d/rc.local: line 12: [: missing `]'`

lzap:
Ok except the ssh, it works fine. Please fix and amend into a single commit and I will retest multiple times to see if race conditions are really gone (so far so good). Looks great!

root/usr/share/fdi/commonfunc.sh
schiermi: When executed this looks better :-)

aschiermeier@pf011e5k:~/Source/foreman-discovery-image> ./root/usr/share/fdi/commonfunc.sh 
This file is meant to be sourced not to be executed: ./root/usr/share/fdi/commonfunc.sh
aschiermeier@pf011e5k:~/Source/foreman-discovery-image>

root/etc/rc.d/rc.local
schiermi: I'm going to fix this. Sorry for not testing SSH access.

root/usr/bin/discovery-fetch-extensions
schiermi: As long as the script (and its includes/requires) do not alter the systemd set environment variables /etc/default/discovery seems to be pretty static and it should be save to drop these statements.

Do we still need ruby for this script afterwards? If not I suggest:

    #!/bin/bash
    
    # Use a commandline option, fdi.zips=zip1.zip,zip2.zip,etc and
    # get the zips from the TFTP server. Try zipserver passed on kernel
    # cmdline first and fall back to DHCP next-server passed in through
    # ENV
    
    local extdir="/opt/extension"
    
    source /usr/share/fdi/commonfunc.sh
    exportKCL
    
    for zip in ${FDI_ZIPS//,/ }
    do
      local tempfile="$(mktemp)"
      curl --fail --silent "tftp://${ZIP_SERVER}/${zip}" -o "${tempfile}"
      local curl_rc="$?"
      if [ "${curl_rc}" -ne 0 ]
      then
        echo "CURL error downloading ${zip} via tftp: ${curl_rc}"
        echo "Problem downloading zipfile ${zip}" >&2
      else
        unzip "${tempfile}" -d "${extdir}"
      fi
      rm "${tempfile}"
    done

lzap:
Almost there, I see a parsing error during fetch script startup:

```
[root@fdi ~]# systemctl status discovery-fetch-extensions.service -l
discovery-fetch-extensions.service - Download zip extensions to this image
   Loaded: loaded (/etc/systemd/system/discovery-fetch-extensions.service; static)
   Active: active (exited) since Mon 2015-08-03 14:11:59 UTC; 3min 39s ago
  Process: 738 ExecStart=/usr/bin/discovery-fetch-extensions (code=exited, status=0/SUCCESS)
 Main PID: 738 (code=exited, status=0/SUCCESS)
   CGroup: /system.slice/discovery-fetch-extensions.service

Aug 03 14:11:59 fdi systemd[1]: Starting Download zip extensions to this image...
Aug 03 14:11:59 fdi discovery-fetch-extensions[738]: /usr/bin/discovery-fetch-extensions: line 8: local: can only be used in a function
Aug 03 14:11:59 fdi systemd[1]: Started Download zip extensions to this image.
```

But network seems to be fine.

root/usr/bin/discovery-fetch-extensions
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# Use a commandline option, fdi.zips=zip1.zip,zip2.zip,etc and
+# get the zips from the TFTP server. Try zipserver passed on kernel
+# cmdline first and fall back to DHCP next-server passed in through
+# ENV
+
+local extdir="/opt/extension"
```
lzap: This is the line.

schiermi: Sorry; transparent proxy gave me an outdated fdi-image.tar during download to my test system.

schiermi:
Please wait; I re-test the whole functionality.

schiermi:
Tested on KVM & HW.

22-discovery.ks
```diff
@@ -30,10 +30,7 @@ systemctl enable nm-prepare.service
 echo " * enabling required system services"
 systemctl enable ipmi.service
 systemctl enable foreman-proxy.service
-systemctl enable discovery-setup.path
-systemctl enable discovery-setup.service
-systemctl enable discovery-autostart.service
-systemctl enable discovery-register.service
```
lzap: All good but discovery-register does not actually start on my instances:

```
[root@fdi ~]# systemctl status discovery-register
discovery-register.service - Register this host in Foreman
   Loaded: loaded (/etc/systemd/system/discovery-register.service; disabled)
   Active: inactive (dead)
```

I will merge this because I am currently working on TUI support which actually starts this service manually after TUI splash screen appears.

lzap: Hmmm setup and autostart are gone/renamed, but I still see usage here:

```
root/etc/systemd/system/discovery-start-extensions.service
4:After=basic.target network.target network-online.target nss-lookup.target discovery-setup.service

root/etc/systemd/system/discovery-register.service
4:After=basic.target network.target network-online.target nss-lookup.target discovery-autostart.service foreman-proxy.service
```

Can you delete these and also make sure register service starts?

lzap:
Merged as 76185479c79860f8dfe040fa50757ff384ebc1b5, thank you!

lzap:
FYI found one additional issue, not sure why I see it now but TFTP download via CURL does not work because firewall blocks client to send ack packets. This enables it:

    echo -e "ip_conntrack_tftp\nnf_conntrack_netbios_ns" > /etc/modules-load.d/tftp-firewall.conf

schiermi:
This is strange. I rebuilt the FDI (fdi-centos7.ks) from a fresh clone. The TFTP download succeeded without any hickups. ip_conntrack_tftp is not loaded in the kernel of the running FDI. iptables -L shows no active rules.

Where is this firewall activated at your side?

I switched to curl to reduce the needed packages and to be able to implement HTTP downloads easily in the future for higher throughput.

lzap:
> Where is this firewall activated at your side?

Sorry, I realized that I actually activated firewalld in my patch. So it
works fine...

-- 
Later,
 Lukas #lzap Zapletal

PR #21, opened 2015-08-07
Fixes #10299 - PXELess discovery with TUI interface

lzap:
The code is done, *feel free to review and test now*.

See the README change for instructions how this is supposed to work.

**Pull requests to merge with this**:

* [x] Discovery plugin: https://github.com/theforeman/foreman_discovery/pull/210
* [x] Smart-proxy discovery image (new) plugin: https://github.com/theforeman/smart_proxy_discovery_image/pull/1
* [ ] Smart-proxy discovery "proxy": https://github.com/theforeman/smart_proxy_discovery/pull/8 (not blocking)
* [ ] Documentation: https://github.com/theforeman/theforeman.org/pull/421 (not blocking)

Instructions:

Build image with this patch and make sure you provide explicit version to it `3.0.0`. This is important, if you build with latest tag `2.1.1` it will not work. Give it any release number you want like 1 or 99.

     ./build-livecd fdi-centos7.ks 99 3.0.0 && sudo ./build-livecd-root 3.0.0

Version can be checked on the TUI interface, in the `/usr/share/fdi/VERSION` file or uploaded facts.

Then checkout foreman develop with foreman_discovery with the patch above. Checkout smart-proxy with discovery "proxy" plugin enabled and with the patch above (note the difference between discovery image and discovery proxy plugins).

The last remaining bit is to install smart_proxy_discovery into the image. You need to do it via ssh since there is no RPM package yet: `gem build smart_proxy_discovery_image.gemspec && sshpass -p redhat scp smart_proxy_discovery_image-1.0.0.gem root@192.168.122.68: && sshpass -p redhat ssh root@192.168.122.68 "gem install smart_proxy_discovery_image-1.0.0.gem && echo 'gem \"smart_proxy_discovery_image\"' > /usr/share/foreman-proxy/bundler.d/discovery_image.rb && systemctl restart foreman-proxy && tail -f /tmp/proxy.log"`

Known issues (TODO):

- [x] better semi-automatic mode (currently it is all or nothing)
- [ ] document how to modify kernel command line options ISO file
- [x] better taxonomy support for discovered hosts (https://github.com/theforeman/foreman_discovery/pull/211)
- [x] better screen texts and buttons
- [x] do IP address CIDR validation on the network form
- [x] move the README block in the official docs
- [x] test with DHCP (but no PXE)
- [x] test without DHCP with multiple NICs
- [x] added extra "append" fact into the template to workaround above

When testing this, do all three scenarios:

* DHCP & PXE (traditional discovery, interrupt with a key)
* DHCP only
* no DHCP


lzap:
@ohadlevy @stbenjam CCing you guys as I expect you to be the revieviers. Discovery plugin part not yet done. Send me your initial thoughts.

lzap:
Actually you can give this a try, but instead of kexec discovery will just reboot the image. Everyting else should just work.

22-discovery.ks
```diff
@@ -7,6 +7,9 @@ yum -y install fedora-release centos-release redhat-release-server || \
 echo " * disabling legacy network services (needed for RHEL 7.0)"
 systemctl disable network.service
 
+echo " * disabling kdump crash service"
+systemctl disable kdump.service
```
lzap: Found this useless (generates dumps in journal).

22-discovery.ks
```diff
@@ -31,6 +37,16 @@ echo " * enabling required system services"
...
+systemctl enable discovery-menu.service
+
+# register service is started manually from discovery-menu
+systemctl disable discovery-register.service
+
+echo " * disabling some unused system services"
+systemctl disable ipmi.service
+
+echo " * open foreman-proxy port via firewalld"
+firewall-offline-cmd --zone=public --add-port=8443/tcp
```
lzap: We were not running firewall by default, this enables it (also need to load TFTP modules above).

25-minimize.ks
```diff
@@ -31,9 +31,7 @@ gzip -9 /usr/share/cracklib/pw_dict.pwd
 
 # remove things only needed during the build process
 echo " * purging packages needed only during build"
-rpm -e syslinux mtools acl ebtables firewalld \
-  libselinux-python python-decorator dracut hardlink kpartx \
-  python-slip python-slip-dbus
+rpm -e syslinux mtools acl
```
lzap: Removed removal of firewall and also other utils. Size bumped just 1 MB.

root/usr/bin/discovery-fetch-extensions
```diff
@@ -13,7 +13,9 @@ exportKCL
 for zip in ${KCL_FDI_ZIPS//,/ }
 do
   tempfile="$(mktemp)"
-  curl --fail --silent "tftp://${ZIP_SERVER}/${zip}" -o "${tempfile}"
+  URL="tftp://${ZIP_SERVER}/${zip}"
+  echo "Downloading $URL"
+  curl --fail --silent "$URL" -o "${tempfile}"
```
lzap: This appears in journal, good to see where are we downloading really from.

lzap:
Notes: Some nm-prepare functionality was moved into separate script nm-configure which is executed separately from udev, nm-prepare service and discovery-menu. Some code from discovery-register ruby script moved to discovery.rb ruby library and is shared with discovery-menu. The way we list root files and executable  flags changed - it's no longer a list, but dynamically created (files must have correct permissions).

john156:
:+1:

lzap:
@stbenjam edited the description with small TODO list, this is ready to be reviewed/tested (one blocker issue currently, see the list pls)

22-discovery.ks
ohadlevy: OK, next time in a different PR :)

22-discovery.ks
```diff
@@ -31,45 +37,83 @@ echo " * enabling required system services"
 systemctl enable ipmi.service
 systemctl enable foreman-proxy.service
 systemctl enable discovery-fetch-extensions.path
+systemctl enable discovery-menu.service
+
+# register service is started manually from discovery-menu
+systemctl disable discovery-register.service
+
+echo " * disabling some unused system services"
+systemctl disable ipmi.service
```
ohadlevy: don't we use it to grab facts from ipmi device?

22-discovery.ks
```diff
@@ -31,45 +37,83 @@ echo " * enabling required system services"
...
+
+echo " * open foreman-proxy port via firewalld"
+firewall-offline-cmd --zone=public --add-port=8443/tcp --add-port=8448/tcp
```
ohadlevy: whats 8448? do we ever use it? should we open 22 for debug too?

22-discovery.ks
```diff
@@ -31,45 +37,83 @@ echo " * enabling required system services"
...
+# certificate is generated by /usr/bin/generate-proxy-cert
+:ssl_certificate: /etc/foreman-proxy/cert.pem
+:ssl_ca_file: /etc/foreman-proxy/cert.pem
+:ssl_private_key: /etc/foreman-proxy/key.pem
+
+:daemon: true
+:http_port: 8448
+:https_port: 8443
+
+# temporary until https://github.com/theforeman/smart-proxy/pull/311 is fixed
```
ohadlevy: https://github.com/theforeman/smart-proxy/pull/311 is merged already

25-minimize.ks
ohadlevy: again, different PR please.

README.md
```diff
@@ -49,7 +50,7 @@ LABEL discovery
 MENU LABEL Foreman Discovery Image
 MENU DEFAULT
 KERNEL boot/fdi-image/vmlinuz0
-APPEND initrd=boot/fdi-image/initrd0.img rootflags=loop root=live:/fdi.iso rootfstype=auto ro rd.live.image acpi=force rd.luks=0 rd.md=0 rd.dm=0 rd.lvm=0 rd.bootif=0 rd.neednet=0 nomodeset proxy.url=http://YOURPROXY proxy.type=proxy
+APPEND initrd=boot/fdi-image/initrd0.img rootflags=loop root=live:/fdi.iso rootfstype=auto ro rd.live.image rd.debug=1 acpi=force rd.luks=0 rd.md=0 rd.dm=0 rd.lvm=0 rd.bootif=0 rd.neednet=0 nomodeset proxy.url=http://YOURPROXY proxy.type=proxy
```
ohadlevy: debug is now default ?

README.md
```diff
@@ -170,6 +119,12 @@ this is mapped to memory. This is intentional, so make sure you have enough
 RAM or you can experience some swapping. Alternatively, change the temp
 directory in the scripts.
 
+It is possible to modify SYSLINUX kernel command line by changing
+livecd-creator code in /usr/lib/python2.7/site-packages/imgcreate/live.py
```
ohadlevy: maybe state minimum versions (or versions that this is possible in?)

root/usr/bin/discovery-debug
```diff
@@ -28,6 +28,9 @@ ip addr show
 echo "* ROUTES *"
 ip route
 
+echo "* NEIGHBOURS *"
```
ohadlevy: interesting, maybe we should considering adding this kind of info via facts too (displaying debug on the UI would be useful too)

root/usr/bin/discovery-menu
```diff
@@ -0,0 +1,576 @@
...
+require "newt"
+require "discovery"
+require "facter"
+require "ipaddr"
+
+# disable GC until https://github.com/theforeman/ruby-newt/issues/2 is fixed
+GC.disable
+
+def fdi_version file = 'VERSION'
+  IO.read("/usr/share/fdi/#{file}").chomp rescue 'GIT'
```
ohadlevy: rescue should not be encouraged, please check if the file exists.
rescue on the entire method as last option.

root/usr/bin/discovery-menu
```diff
@@ -0,0 +1,576 @@
...
+
+def enable_root_account(pass)
+  command("echo 'root:#{pass}' | chpasswd && systemctl restart sshd.service")
+end
+
+def error_box(msg, debug_msg = nil)
+  log_err msg
+  ex_msg = debug_msg.to_s
+  if debug_msg.is_a? Exception
+    ex_bt = (debug_msg.to_s + "\n" + debug_msg.backtrace.join("\n")) rescue 'N/A'
```
ohadlevy: please use meaningful variable names.

root/usr/bin/discovery-menu
```diff
@@ -0,0 +1,576 @@
...
+    log_err ex_bt
+  end
+  log_err ex_msg
+  Newt::Screen.centered_window(74, 20, "Fatal error")
+  f = Newt::Form.new
+  t_desc = Newt::Textbox.new(1, 1, 70, 13, Newt::FLAG_SCROLL)
+  random_pass = (1...10).map { (65 + rand(26)).chr }.join
+  t_desc.set_text "#{msg}:\n#{ex_msg}\n\n#{ex_bt}\n\n" +
```
ohadlevy: consider extracting to a separate method, or maybe even erb?

root/usr/bin/discovery-menu
```diff
@@ -0,0 +1,576 @@
...
+  f.add(b_ok, t_desc)
+  f.run
+  enable_root_account(random_pass)
+  exit(1)
+end
+
+def command(cmd)
+  log_msg "TUI executing: #{cmd}"
+  # do not run real commands in development (non-image) environment
+  return unless File.exist? '/usr/share/fdi/VERSION'
```
ohadlevy: please reuse the method that checks the version from above.

root/usr/bin/discovery-menu
```diff
@@ -0,0 +1,576 @@
...
+  output = `#{cmd} 2>&1`
+  if $? != 0
+    error_box("Command failed: #{cmd}", output)
```
ohadlevy: should you return here or always provide the output from the method output?

root/usr/bin/discovery-menu
```diff
@@ -0,0 +1,576 @@
...
+  t.set_text(text_help)
+
+  l_press = Newt::Label.new(-1, -1, "< Press any key (10) >")
+
+  main_grid = Newt::Grid.new(1, 2)
+  but_grid = Newt::Grid.new(1, 1)
+
+  but_grid.set_field(0, 0, Newt::GRID_COMPONENT, l_press, 0, 0, 0, 0, 0, 0)
+
+  main_grid.set_field(0, 0, Newt::GRID_COMPONENT, t, 0, 0, 0, 1, 0, 0)
```
ohadlevy: some descriptions could be nice :)

root/usr/bin/discovery-menu
```diff
@@ -0,0 +1,576 @@
...
+def screen_welcome
+  text_help, tw, th = Newt.reflow_text(<<EOT, 60, 5, 5)
+Select Manual network setup to select primary interface, configure network (no DHCP required), \
+setup Foreman server credentials, add custom facts and trigger auto-provisioning \
+via Discovery rules. This will lead to kernel reload (kexec) into installer. \
+Select Discover with DHCP to select primary interface and proceed with DHCP configuration \
+and standard discovery without any custom facts. This will reboot the host once the system \
+is provisioned either manually or via Discovery rules.
+EOT
+  t = Newt::Textbox.new(-1, -1, tw, th, Newt::FLAG_WRAP)
```
ohadlevy: variable names..

ohadlevy:
1. requires testing, i only did a very brief code review
2. I really dislike the tui in one large file, it feels it needs to be re factored to multiple classes, and missing tests.

22-discovery.ks
lzap: We do but the service is not involved AFAIK. @stbenjam can you test this when doing the ipmi patch please?

22-discovery.ks
lzap: 8448 http. 8443 https, good point ssh, need to enable it in the TUI

25-minimize.ks
lzap: Yeah this one is long-term patch. Improved things here and there.

22-discovery.ks
lzap: Wait, ssh port is enabled by default, no need to add it explicitly (Anaconda does that by default).

22-discovery.ks
lzap: Yeah, right. Fixed, finally smart-proxy works with syslog/systemd!

README.md
lzap: Yes, it will basically give you emergency shell instead of halting system when something bad happens. Better than halt... You can still investigate logs.

lzap: FYI this is a Dracut feature not ours.

README.md
lzap: I have a TODO item to expand this. Existing image should be modifiable, don't know how yet. Will rewrite this section.

root/usr/bin/discovery-debug
lzap: Hmmm indeed, but what for? I'd rather not report things we don't process. I added this because I sometimes mis-start a VM with multiple NICs on the same network.

root/usr/bin/discovery-menu
lzap: Nitpicker! :-)

root/usr/bin/discovery-menu
lzap: Hmmm I don't know, we have more of these.

root/usr/bin/discovery-menu
lzap: Yeah sure.

root/usr/bin/discovery-menu
lzap: Error box method is a fatal one, ends with `exit(1)`. This will lead to systemd restarting the service after some time.

root/usr/bin/discovery-menu
lzap: Descriptions?

lzap:
Thanks for review, pushing the fixes. This really need bunch of testing indeed. And I can split into multiple files - one per screen perhaps.

Not sure how to test the TUI. Do you mean to write unit tests for the few non-screen methods that spawns some commands? Isn't this an overkill?

lzap:
I have split the discovery-menu into smaller files.

root/usr/bin/discovery-menu
```diff
@@ -0,0 +1,17 @@
...
+require 'discovery/screen/foreman'
+require 'discovery/screen/network'
+require 'discovery/screen/primary_iface'
+require 'discovery/screen/ssh'
+require 'discovery/screen/status'
+require 'discovery/screen/welcome'
+
+# Disable GC until https://github.com/theforeman/ruby-newt/issues/2 is fixed,
+# see also: http://projects.theforeman.org/issues/11623
+GC.disable
```
lzap: For the record, there was a problem in the newt cleanup code which I have resolved in https://github.com/theforeman/ruby-newt/commit/5e0144a5dceba285a8d4a516086e0045087294dd but until I will have a chance to do full code review in regard to memory leaks and GC problems, I'd rather keep this disabled for 3.0.0 release. Afterwards let's enable it for nightly builds.

I am tracking this in http://projects.theforeman.org/issues/11623

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
```diff
@@ -0,0 +1,46 @@
+def screen_foreman mac = nil, gw = nil, proxy_url = cmdline('proxy.url'), proxy_type = cmdline('proxy.type')
+  Newt::Screen.centered_window(59, 20, "Foreman credentials")
+  f = Newt::Form.new
+  t_desc = Newt::Textbox.new(2, 2, 54, 6, Newt::FLAG_WRAP)
+  t_desc.set_text "Provide full URL (http(s)://host:PORT) to the Server or Proxy " +
+  "according to the type selected. Ports are usually 8443, 8448 or 9090 according " +
+  "to configuration."
```
stbenjam: Include 443 to cover Foreman Server connection

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
```diff
@@ -0,0 +1,46 @@
...
+  l_type = Newt::Label.new(2, 10, "Connection type:")
+  t_url = Newt::Entry.new(20, 8, "", 36, Newt::FLAG_SCROLL)
+  r_server = Newt::RadioButton.new(20, 10, "Server", 0, nil)
+  r_proxy = Newt::RadioButton.new(32, 10, "Proxy", 1, r_server)
+  b_ok = Newt::Button.new(34, 15, "Next")
+  b_cancel = Newt::Button.new(46, 15, "Cancel")
+  if gw.nil?
+    proxy_url ||= ''
+  else
+    proxy_url ||= "https://#{gw}:8443"
```
stbenjam: Can the default just be 443? The most common case is probably directly the Foreman.  And 8443 doesn't work for Katello, so that might be confusing for users.  Let them enter the proxy port, if they want to connect to one

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+  proxy_type ||= 'proxy'
```
stbenjam: Ditto, default to server IMHO

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
lzap: > Can the default just be 443? The most common case is probably directly the Foreman.  And 8443 doesn't work for Katello, so that might be confusing for users.  Let them enter the proxy port, if they want to connect to one

Default behavior of discovery is via proxy. We actually require the port
number to be present, this switch actually changes the way how the
communication is done (proxy has a different API than Foreman).

-- 
Later,
 Lukas #lzap Zapletal

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
lzap: > Ditto, default to server IMHO

I try to align with our recommended setup, which is proxy variant. This
is also the default behavior.

I understand this can be confusing downstream, I can introduce some
DEFAULT_PORT option perhaps for the build phase?

-- 
Later,
 Lukas #lzap Zapletal

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
stbenjam: > Default behavior of discovery is via proxy.

Why?

> We actually require the port
number to be present, this switch actually changes the way how the
communication is done (proxy has a different API than Foreman).

The default port being 8443 is going to be massively confusing to Katello/Satellite users.

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
stbenjam: > I understand this can be confusing downstream, I can introduce some DEFAULT_PORT option perhaps for the build phase?

Ah just saw this, yea that would be OK, although I still question the behavior of operating in proxy mode by default.  Everyone will have a Foreman; not everyone has a proxy.

lzap: With discovery proxy is required, PXE does not operate well on WAN networks.

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
lzap: I think we can do some branding here.

lzap:
Amended the changes, also updated the kexec template, if you already checked out, update it on your instance. Solved the issue with multiple NICs.

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
stbenjam: >  With discovery proxy is required, PXE does not operate well on WAN networks.

I really don't understand what you mean, what does WAN have to do anything?

A *default* Foreman install is standalone and fully functional, deploying discovery should be easy.  There's no reason to force the standalone users to install the smart_proxy_discovery package or change all the configuration to use the foreman type.

stbenjam:
Hey sorry its taken so long to get to look at this.  I really like the TUI interface and workflow :+1: 

1. How is DHCP but no PXE supposed to work? It just stays here after selecting DHCP:

nopxe

2. Manual configuration works up until kexec - is there any way to test kexec?

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
lzap: Yeah, I understand what you mean. I was trying to be compliant with http://theforeman.org/plugins/foreman_discovery/4.0/index.html#3.1ForemanDiscoveryplugin but it looks like it's better to have Foreman w/o proxy as the default value. Changing this and also our docs in the separate PR.

lzap:
1) Click on "Discover with DHCP", select the primary interface to configure and then this screen appears. We do this:

```
      configure_network false, primary_mac
      disable_tftp_extensions
      start_discovery_service
      :screen_status      
```

This configures network via NM, disables TFTP extensions service and starts discover-host systemd service. And then we go into the status screen which waits 10 seconds and then refresh the screen. If there was no status provided, investigate jornal what happened, You should see discover-host entries there. When request is processed, it puts a file in /tmp folder and TUI picks it up to show it on screen (I wanted to avoid IPC here). Try to click on "Status" button several times to reload. Maybe a race condition or something? Slow network?

This 10 second excersise is because limitation of newt - once you start a form, there is no way to refresh the screen until user submits the screen. So I wait 10 seconds and then do initial draw of the form. You can refresh with Status button. See `generate_info` method.

I am adding extra text there to notify the user about this in this form: `status = "N/A (use Status to update)"`.

2) What do you see in foreman log? Make sure you install the `smart_proxy_discovery_image` rubygem onto the running image (see the long command in the summary). If you dont do that, then foreman fails to call the new power_api which is not yet available which fails with 404 or 500.

root/usr/lib64/ruby/vendor_ruby/discovery/screen/foreman.rb
stbenjam: It's not urgent, as I supposed it is out of the scope of this PR, but just because there's a lot more proxy-default code here I thought I'd mention it.

lzap:
In regard to the refreshing screen, I hooked HUP signal to do screen refresh. But this did not work as I am not able to do many things in trap handler. By default I just call refresh (I'd need to change actually some data).

Sorry I just realized in DHCP workflow there has little bit different implementation. Facts upload is performed directly, so TUI is blocked for a moment and then status screen is showed without any countdown magic. Investigate what happened.

stbenjam:
> This configures network via NM, disables TFTP extensions service and starts discover-host systemd service. And then we go into the status screen which waits 10 seconds and then refresh the screen. 

In DHCP mode, shouldn't I be able to enter the foreman/proxy server details?  It goes straight to discovery status

22-discovery.ks
stbenjam: Yup, service is not required

stbenjam:
> 2) What do you see in foreman log? Make sure you install the smart_proxy_discovery_image rubygem onto the running image (see the long command in the summary). If you dont do that, then foreman fails to call the new power_api which is not yet available which fails with 404 or 500.

I think I have everything up to date. The gem is installed on the host, and discovery_image is loaded  in Foreman proxy.

I see this in the logs:


```
2015-09-04T13:43:18 [app] [W] Unable to reboot mac52540033ed00
 | ArgumentError: Legacy direct service only supports http scheme
 | /home/vagrant/foreman_discovery/app/services/foreman_discovery/node_api/power.rb:37:in `initialize'
 | /home/vagrant/foreman_discovery/app/services/foreman_discovery/node_api/power.rb:9:in `new'
 | /home/vagrant/foreman_discovery/app/services/foreman_discovery/node_api/power.rb:9:in `legacy_direct_
service'
 | /home/vagrant/foreman_discovery/app/models/host/discovered.rb:147:in `reboot'
 | /home/vagrant/foreman_discovery/app/models/host/managed_extensions.rb:28:in `setReboot'
 | /home/vagrant/foreman/app/models/concerns/orchestration.rb:152:in `execute'
 | /home/vagrant/foreman/app/models/concerns/orchestration.rb:97:in `block in process'
 | /home/vagrant/foreman/app/models/concerns/orchestration.rb:89:in `each'
 | /home/vagrant/foreman/app/models/concerns/orchestration.rb:89:in `process'
 | /home/vagrant/foreman/app/models/concerns/orchestration.rb:23:in `post_commit'
```

lzap:
> I think I have everything up to date. The gem is installed on the host, and discovery_image is loaded  in Foreman proxy.

Have you provided the version number during building? We check fdi
version and if < 3.0.0 then we fallback to legacy mode. This is because
after upgrade users need to reboot into new images...

-- 
Later,
 Lukas #lzap Zapletal

lzap:
> In DHCP mode, shouldn't I be able to enter the foreman/proxy server details?  It goes straight to discovery status

Good point, forgot about it. Will add.

-- 
Later,
 Lukas #lzap Zapletal

lzap:
> In DHCP mode, shouldn't I be able to enter the foreman/proxy server details?  It goes straight to discovery status

Fixed!

Thanks for review, appreciated. Quite big feature, if there are no
blockers I'd prefer to merge it as our QA will carry on testing this. I
can fix reported bugs as we go.

-- 
Later,
 Lukas #lzap Zapletal

sghai:
Yeah.. happy to test this once it will be available via nightly.  Thanks

stbenjam:
> Thanks for review, appreciated. Quite big feature, if there are no blockers I'd prefer to merge it as our QA will carry on testing this. I can fix reported bugs as we go.

- [ ] Merged https://github.com/theforeman/smart_proxy_discovery_image/pull/1 - can you release to rubygems and build a package?
- [ ] https://github.com/theforeman/foreman_discovery/pull/210 needs to be green.

Then I think we can merge this one, and start getting feedback from QE

lzap:
@stbenjam rubygem-smart_proxy_discovery_image-1.0.4-1.el7.noarch.rpm hits the nightlies, please pull and rebuild the image, it now works out-of-box.

lzap:
@stbenjam Ori is away until Wed, unit tests are blocked because develop currently fails, this fixes it: https://github.com/theforeman/foreman_discovery/pull/214 (removed unused RABL templates and enabled some skipped tests)

lzap:
@stbenjam edited summary with checklist what needs to be merged, still working on the proxy part but this does not block us from testing this with direct node-foreman setting, thanks!

lzap:
@stbenjam done with both remaining unit test failures, covered with lots of tests. /me crosses his fingers.

stbenjam:
Looks good! Merged as f8d1accb8e3819cb47beed49f408abb8c34c3c4b, the remaining non-blocking items can be done elsewhere.

PR #23, opened 2015-09-04
Do not cache initial 10 facter runs on discovered node

koweblomke:
start discovery-register service after network-online
because interfaces have to be renamed before starting the service.

because facter caches the facts it would otherwise hold the initial interface names (ethx). And the service keeps sending these (wrong) interface names. 

This causes foreman to be unable to discover the host with (Debug) message: "Unable to assign subnet, primary interface is missing IP address"

### DIFF
--- a/root/usr/bin/discovery-register
+++ b/root/usr/bin/discovery-register
@@ -84,9 +84,15 @@ facts = Hash[Facter.to_hash.select {|k,v| k =~ /address|hardware|manufacturer|pr
 facts.keys.sort.each {|k| log_msg " #{k}: #{facts[k]}"}
 
 # loop, but only upload on changes
+i = cmdline("fdi.cachefacts", 10).to_i
 while true do
   uninteresting_facts=/kernel|operatingsystem|osfamily|ruby|path|time|swap|free|filesystem|version|selinux/i
   facts = Facter.to_hash.reject! {|k,v| k =~ uninteresting_facts }
+  if i > 0
+    # clear cache because during bootproces facts may change
+    Facter.clear
+    i -= 1
+  end
   unless YAML.load(read_cache) == facts
     log_msg "Fact cache invalid, reloading to foreman"
     write_cache(YAML.dump(facts)) if upload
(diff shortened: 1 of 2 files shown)


lzap:
This is indeed problematic, systemd is very fast in spawning services. What is weird is that this affects you because this kind of service is type of `idle`.  http://www.freedesktop.org/software/systemd/man/systemd.service.html

I am not against this change, I just doubt it fixes the issue.

But your idea with not caching the facts is interesting, would you mind changing the code in a way that initial 10 runs are *not* cached (while the rest is cached)? I dont want idling servers to be working hard but it is fine for initial few minutes IMHO.

lzap:
The motivation behind cache is becuse facter is CPU heavy, I am fine with 10 initial runs without cache tho.

koweblomke:
Facter cache now cleared first 10 times.

We have also discovered that changes since 2.1 stable had a positive influence on our issue.
probably commit https://github.com/theforeman/foreman-discovery-image/commit/76185479c79860f8dfe040fa50757ff384ebc1b5 has already fixed our problem.

lzap:
Cool, would you mind to refactor it a bit and read the kernel command line option with `cmdline("fdi.cachefacts", 10)` instead of hardcoding the number? The helper is defined in `discovery.rb` file. Users can override this.

root/usr/bin/discovery-register
```diff
@@ -84,9 +84,15 @@ facts = Hash[Facter.to_hash.select {|k,v| k =~ /address|hardware|manufacturer|pr
 facts.keys.sort.each {|k| log_msg " #{k}: #{facts[k]}"}
 
 # loop, but only upload on changes
+i = 10
 while true do
   uninteresting_facts=/kernel|operatingsystem|osfamily|ruby|path|time|swap|free|filesystem|version|selinux/i
   facts = Facter.to_hash.reject! {|k,v| k =~ uninteresting_facts }
+  unless i == 0
```
lzap: Pls do `if i > 0` instead which is safer.

lzap:
And please squash into single commit: `Fixes #11678 - Do not cache initial 10 facter runs on discovered node".

lzap:
Thanks, I still see five commits instead of one: https://github.com/theforeman/foreman-discovery-image/pull/23/commits Then I can merge!

lzap:
Ping, we are approaching image 3.0.0 rc1 build, I'd like to have it there.

lzap:
Merged as 153c964ca9496d04f21541895677d0c7930c81b9, thank you!

PR #24, opened 2015-09-18
Add Newt::FLAG_SCROLL to allow facts name and value bigger than field length

gloppasglop:
Add Newt::FLAG_SCROLL to allow facts name and value bigger than field length


lzap:
Thanks! Any feedback on the new image nightly build will be highly appreciated!

lzap:
Merged as 20547b9f4b96602b6b2957e8a8fee248e2c3abf7, thank you!

PR #27, opened 2015-10-01
Fixes #11932 - escape messages for syslog

lzap:
It looks like syslog package insist on formatted strings.


lzap:
Also errors during logging must not be fatal.

stbenjam:
Merged as 67b4ef645ef2ce95890a484000fbce186ba0ce9b

PR #29, opened 2015-10-03
Enable LLDP facts

jb11211:
This change enables the lldpad daemon, turns on lldp for
all interfaces (excluding lo) and then uses the openlldp.rb
written by razorsedge to report the lldp data in facter

Credit where it's due. the openlldp.rb file is from here:
https://github.com/razorsedge/puppet-openlldp/blob/master/lib/facter/openlldp.rb

### DIFF
--- a/root/usr/lib64/ruby/vendor_ruby/discovery.rb
+++ b/root/usr/lib64/ruby/vendor_ruby/discovery.rb
@@ -93,7 +93,9 @@ def proxy_type
 end
 
 def write_tui result = 'success', code, body
-  File.open("/tmp/discovery-http-#{result}", 'w') do |file|
+  filename = "/tmp/discovery-http-#{result}"
+  log_debug "Wrote result #{code} to #{filename}"
+  File.open(filename, 'w') do |file|
     file.write("#{code}: #{body}")
   end
 end
@@ -101,6 +103,10 @@ end
 def upload(uri = discover_server, type = proxy_type, custom_facts = {})
   unless uri
     log_err "Could not determine Foreman instance, add foreman.url or proxy.url kernel command parameter"
+    return
+  end
+  unless uri.is_a? URI
+    log_err "Upload#uri must be type of URI"
     return
   end
   if uri.host.nil? or uri.port.nil?
--- a/root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
+++ b/root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
@@ -118,6 +118,19 @@ def cleanup
   Newt::Screen.finish
 end
 
+def detect_first_nic_with_link
+  log_debug "Traing to guess the first NICs with link, fdi.pxmac was NOT provided"
+  mac = ''
+  Dir.glob('/sys/class/net/*') do |ifn|
+    name = File.basename ifn
+    next if name == "lo"
+    mac = File.read("#{ifn}/address").chomp rescue "??:??:??:??:??:??"
+    link = File.read("#{ifn}/carrier").chomp == "1" rescue false
+    break if link
+  end
+  mac
+end
+
 log_msg "Kernel opts: #{cmdline}"
 
 def main_loop
@@ -130,42 +143,58 @@ def main_loop
   Newt::Screen.new
   Newt::Screen.push_helpline("Foreman Discovery Image v#{fdi_version} (#{fdi_release})")
 
-  if cmdline('fdi.pxauto')
-    # automated pxeless discovery via kernel command line options
-    mac = cmdline('fdi.pxmac')
-    ip = cmdline('fdi.pxip')
-    gw = cmdline('fdi.pxgw')
-    dns = cmdline('fdi.pxdns')
-    configure_network true, mac, ip, gw, dns
-    proxy_url = cmdline('proxy.url')
-    proxy_url = cmdline('proxy.type')
-    facts = new_custom_facts(mac)
-    [1..9].each do |n|
-      if (fact_name = cmdline("fdi.pxfactname#{n}"))
-        fact_value = cmdline("fdi.pxfactvalue#{n}")
-        facts[fact_name] = fact_value
-      end
-    end
-    if perform_upload(proxy_url, proxy_type, custom_facts)
-      [:screen_status, generate_info('AWAITING KEXEC INTO INSTALLER')]
-    else
-      error_box("Unable to send facts", "Automated fact upload failed")
-    end
-  end
-
   if cmdline('BOOTIF')
-    func = :screen_countdown
+    # Booted via PXE
+    active_screen = :screen_countdown
   else
-    # the image was booted from ISO directly
-    func = :screen_welcome
-  end
-  while ! [:quit].include? func
-    if func.is_a?(Array)
-      log_debug "Entering #{func[0]}"
-      func = send(*func)
+    # Booted from ISO
+    if cmdline('fdi.pxauto')
+      # Unattended PXE-less provisioning
+      log_debug "Unattended provisioning started"
+      proxy_url = cmdline('proxy.url') || error_box("Option proxy.url was not provided, cannot continue")
+      proxy_url = URI.parse(proxy_url) rescue error_box("Unable to parse proxy.url URI: #{proxy_url}")
+      proxy_type = cmdline('proxy.type') || error_box("Option proxy.type was not provided, cannot continue")
+      log_debug "proxy.url=#{proxy_url} proxy.type=#{proxy_type}"
+      mac = cmdline('fdi.pxmac', detect_first_nic_with_link)
+      ip = cmdline('fdi.pxip') || error_box("Option fdi.pxip was not provided, cannot continue")
+      gw = cmdline('fdi.pxgw') || error_box("Option fdi.pxgw was not provided, cannot continue")
+      dns = cmdline('fdi.pxdns')
+      log_debug "fdi.px: mac=#{mac} ip=#{ip} gw=#{gw} dns=#{dns}"
+      action = Proc.new do
+        log_debug "Unattended network configuration started"
+        status = configure_network true, mac, ip, gw, dns
+        log_debug "Unattended network configuration finished, result: #{status}"
+        log_debug "Unattended facts collection started"
+        facts = new_custom_facts(mac)
+        [1..9].each do |n|
+          if (fact_name = cmdline("fdi.pxfactname#{n}"))
+            fact_value = cmdline("fdi.pxfactvalue#{n}")
+            facts[fact_name] = fact_value
+          end
+        end
+        log_debug "Unattended facts collection finished"
+        log_debug "Unattended facts upload started"
+        result = perform_upload(proxy_url, proxy_type, facts)
+        log_debug "Unattended facts upload finished, result: #{result}"
+        result
+      end
+      active_screen = [:screen_info, action,
+        "Performing unattended provisioning via NIC #{mac} with IP #{ip} (gw #{gw} DNS #{dns}) sending facts to #{proxy_url} of endpoint type #{proxy_type}...",
+        "Unattended provisioning failed: unable to upload facts",
+        [:screen_status, generate_info('Unattended fact upload OK - awaiting kexec')],
+        [:screen_status, generate_info('Unattended fact upload FAILED')]]
     else
-      log_debug "Entering #{func}"
-      func = send(func)
+      # Attended PXE-less provisioning
+      active_screen = :screen_welcome
+    end
+  end
+  while ! [:quit].include? active_screen
+    if active_screen.is_a?(Array)
+      log_debug "Entering #{active_screen[0]}"
+      active_screen = send(*active_screen)
+    else
+      log_debug "Entering #{active_screen}"
+      active_screen = send(active_screen)
     end
     Newt::Screen.pop_window()
   end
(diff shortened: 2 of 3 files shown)


ares:
wow, this could be great start for Foreman network discovery

jb11211:
thanks @ares . I've just refactored the change into files rather than using cat << EOF .... 

It's the same exact change. just cleaner.

fdi-centos7.ks
```diff
@@ -3,4 +3,5 @@
 %include 20-packages.ks
 %include fdi-install.ks
 %include 22-discovery.ks
+%include 23-lldp.ks
```
lzap: This can be part of `22-discovery.ks` file.

root/usr/share/fdi/facts/openlldp.rb
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  # Remove interfaces that pollute the list (like lo and bond0).
+  Facter.value('interfaces').split(/,/).grep_v(/^lo$|^bond[0-9]/).each do |interface|
+    # Loop through the list of LLDP TLVs that we want to present as facts.
+    lldp.each_pair do |key, value|
+      Facter.add("lldp_neighbor_#{key}_#{interface}") do
+        setcode do
+          result = ""
+          output = Facter::Util::Resolution.exec("lldptool get-tlv -n -i #{interface} -V #{value} 2>/dev/null")
```
ohadlevy: nitpick: I would extract the exec from both the lldp.each_pair and facter.add loops, it can be the first thing you do, and not enter any other code if there is no output.

root/usr/lib/systemd/system/lldpad_enable_all.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Enable LLDP on all interfaces
+After=lldpad.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/lldptool_enable_all.sh
+
+[Install]
+WantedBy=multi-user.target
```
lzap: This will not work correctly, systemd will sometimes spawn this *before* network drivers are even loaded. We faced this kind of issues already. See what we do in `22-discovery.ks` already for network manager configuration via udev:

```
cat > /etc/udev/rules.d/81-nm-prepare.rules <<'UDEV'
ACTION=="add", SUBSYSTEM=="net", NAME!="lo", RUN+="/usr/bin/systemd-cat -t nm-prepare /usr/bin/nm-prepare %k"
UDEV
```

Please run the enable script in a similar way for each particular device (one by one). Also rename it to `enable-lldp` without extension.

root/usr/share/fdi/facts/openlldp.rb
```diff
@@ -0,0 +1,99 @@
...
+#
+
+require 'facter/util/macaddress'
+
+# http://www.ruby-forum.com/topic/3418285#1040695
+module Enumerable
+  def grep_v(cond)
+    select {|x| not cond === x}
+  end
+end
```
lzap: Would you mind *not* extending `Enumerable` here and creating a helper method? This will be unlikely shared and Facter is fragile codebase already.

root/usr/share/fdi/facts/openlldp.rb
```diff
@@ -0,0 +1,99 @@
...
+
+#if Facter::Util::Resolution.which('lldptool')
```
lzap: Leftover?

root/usr/share/fdi/facts/openlldp.rb
```diff
@@ -0,0 +1,99 @@
...
+    'portID'       => '2',
+    'sysName'      => '5',
+    'mngAddr_ipv4' => '8',
+    'mngAddr_ipv6' => '8',
+    'PVID'         => '0x0080c201',
+    'MTU'          => '0x00120f04',
+  }
+
+  # Remove interfaces that pollute the list (like lo and bond0).
+  Facter.value('interfaces').split(/,/).grep_v(/^lo$|^bond[0-9]/).each do |interface|
```
lzap: I'd do `/^lo$|^bond[0-9]+/` perhaps, people have weird setups :-)

root/usr/share/fdi/facts/openlldp.rb
```diff
@@ -0,0 +1,99 @@
...
+          else
+            # No output from lldptool
+            result = nil
+          end
+          result
+        end
+      end
+    end
+  end
+end
```
lzap: Missing EOL

lzap:
Nice contribution, thanks. Couple of items above.

lzap:
@ohadlevy by the way the items above (ordering issue with systemd) will sove the initial fact upload which is missing the values.

jb11211:
hey guys. the openlldp.rb file was copied from another source and I'm reluctant to change it here for two reasons:
1) I don't know enough ruby to do it better.
2) the changes, if appropriate should be contributed upstream.

Perhaps someone who knows ruby better could contribute a fix upstream.

Still thinking about how to deal with the systemd starting services before networking issue.

lzap:
Sorry for the delay, I was working on Discovery 4.1 and the new image build. It's getting out this week, then we can merge this.

lzap:
Ok there is one licensing problem, this repo is GNU GPLv2 while the file you copied is Apache License v2. These are not compatible. I think we can easily upgrade to GNU GPLv3.

jb11211:
Would it be better if I refactored the code to pull down this file during the image build? it's not the most robust solution but it would solve two problems:
1) we would no longer duplicate files developed elsewhere
2) we wouldn't have this license issues (I think)

netflash:
Can I ask a question ? 
I was not able to force lldpAD to work with CDP, so I have to switch to https://vincentbernat.github.io/lldpd/
Am I doing something wrong or lldpAP really can't work with CDP ?

lzap:
No downloading is allowed when we build packages, so this is not preferred. Also I don't like git modules :-)

I will re-license this repo to GNU GPL v3 and then we can include the file. Can you still fix the comments I had? It's fine to change forked code. If you want to push improvements upstream is up to you. Best way would be you to push the changes upstream first and then copy over the final version in here.

jb11211:
Hey Lzap,

I genuinely don't know enough Ruby to fix your comments. anything I would submit would be worse than what this has now.

I'm still yet to find time to work with your systemd hack for networking. not really sure what to do there.

When I do find time I will try and fix that, although I'm yet to hit the issue you described.

beddari:
Should really be discussed elsewhere, but to provide some background info around what @netflash is asking ...

FWIW we've been using https://vincentbernat.github.io/lldpd/ lldpd on our servers for quite some time. This summaries our experience:

* it has easy, user friendly command line tooling
* it works fine with cdp and others
* it reports most information, more is good :)
* it is, from what we can see, the best mantained codebase
* it is the default lldp on e.g Cumulus Linux, Big Switch ...
* Packages on http://software.opensuse.org/download.html?project=home:vbernat&package=lldpd 

http://www.kempgen.net/voip/lldp-agents.html

lzap:
Ok can you add source URL and license information to the top of the `openlldp.rb` file header?

And then add this to README.md before Contribution section:

```
License
-------

The kickstart file, utility scripts and other software in this repo is licensed under GNU GPL v2 or later. Exceptions are individually commented in file headers.

Generated image is covered by additional licenses, refer to Fedora and CentOS licensing information.
```

Or something like that. Then I can start testing this.

korekhov:
+1 for using LLDPD insstead of LLDPAD (aka OpenLLDP) as the latter is mostly Intel's implementation as per http://www.kempgen.net/voip/lldp-agents.html

We also been using LLDPD for years now and pretty happy with it.

beddari:
@korekhov @netflash What we really should do is just suck it up and write some code ;-) I've got it planned but no timeframe yet.

korekhov:
If you mean the code to parse LLDP info, here's the excerpt from my nics.rb script:

      # Get a switch and port info if LLDP is installed
      tool = '/usr/sbin/lldpctl'
      if FileTest.executable?(tool)
        switch = ""
        port = ""
        vlan = ""
        output = Facter::Util::Resolution.exec("#{tool} -f keyvalue #{i} 2>&1")
        if output and not output.empty?
          output.each_line do |l|
            l.strip!
            next if l.empty?
            switch = l.split(/=/,2)[1] if l =~ /chassis.name/
            port = l.split(/=/,2)[1] if l =~ /port.ifname/
          end
        end
        output = Facter::Util::Resolution.exec("#{tool} #{i} 2>&1")
        if output and not output.empty?
          output.each_line do |l|
            l.strip!
            next if l.empty?
            vlan = l.split[1].gsub(/,/,'') if l =~ /pvid: yes/
          end
        end
        if not switch.empty? and not port.empty?
          nics[i]['switch'] = switch
          nics[i]['port'] = port
          nics[i]['vlan'] = vlan
        end
      end

jb11211:
@lzap changes made as requested.

root/usr/lib/systemd/system/enable-lldp
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Enable LLDP on all interfaces
+After=lldpad.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/enable-lldp
+
+[Install]
+WantedBy=multi-user.target
```
lzap: This is missing `.service` extension.

Since you iterate over network ifaces, you need to make sure this service is started after network initialization (drivers are fully loaded), but before registration happens or proxy starts up:

```
After=network.target network-online.target
Before=foreman-proxy.service discovery-register.service
```

root/usr/bin/enable-lldp
```diff
@@ -0,0 +1,23 @@
...
+    lldptool -i $int -T -V $var enableTx=yes
+  done
+
+}
+
+for inter in $(ls /sys/class/net/ | grep -v "lo"); do
+ enable_lldp_int $inter $ipaddress portDesc sysName sysDesc sysCap
+done
+
+exit 0
```
lzap: Retab / reformat this file please.

22-discovery.ks
```diff
@@ -132,4 +132,8 @@ echo "alias 'yum=echo DO NOT USE YUM; yum'" >> /root/.bashrc
 # Base env for extracting zip extensions
 mkdir -p /opt/extension/{bin,lib,lib/ruby,facts}
 
+echo " * setting up lldp service"
+systemctl enable lldpad.service
+systemctl enable enable-lldp
```
lzap: Create this as `.service` file.

lzap:
I prefer the package that is in CentOS itself (not EPEL).

lzap:
Ping - I had couple of other comments, I would like to squeeze this patch in the next release! Thanks.

jb11211:
The Bash script is tabbed/formatted in accordance with this style guide:
http://wiki.bash-hackers.org/scripting/style

Does theforeman have it's own style guide for Bash? or in general?

Other requested changes have been made.

lzap:
It's fine, please squash into one commit.

lzap:
There is one more issue, you have `After=network.target` but that does not mean your service will start after all the network devices are initialized. Therefore this can easily end up with LLDP enabled only on the first (few) devices and not the others.

Please use udev instead, it's easy, you don't need to write a systemd service at all. See this for an example: https://github.com/theforeman/foreman-discovery-image/blob/master/22-discovery.ks#L21-L23

Also squash please.

jb11211:
I don't think it makes any sense to manage service startup out of Udev.

the setup of networking interfaces, sure. that's what udev is for. but trying to use it to ensure start up order in place of the actual startup daemon (systemd) is really not making sense to me.

At your request the enable-lldp service depends on the "network-online.target". here I will quote directly from that targets documentation (https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/):

network-online.target is a target that actively waits until the nework is "up", where the definition of "up" is defined by the network management software. .... It's primary purpose is to actively delay activation of services until the network is set up. .... By default all remote mounts defined in /etc/fstab pull this service in, in order to make sure the network is up before it is attempted to connect to a network share. 

PS, to the ppl saying we should use "LLDPD".  This patch uses lldpad because thats readily available and works :)

lzap:
I've worked close with systemd developers when I was designing `nm-prepare` script which does the similar thing as `enable-llpd` - it configures something for each individual network device. That's exactly what udev was built for! Load firmware, configure device or service hooked to the device.

Network-online target indeed works as you describes, and this is *not* what's needed in this case. It will be fulfilled even when devices for other network cards are *not yet loaded*. I've seen this myself, discovered host was "online" (and responding) while it was still initializing 7th and 8th network driver. That's because NetworkManager considers a host "online" when *at least one of the interfaces acquire DHCP address*.

This is highly anticipated feature and I am really happy for your contribution, I am just trying to make sure it will work properly. Trust me, I've seen *a lot* of discovery startup race conditions in the last two years and it took me days of reading manuals and experimenting to learn how systemd and NetworkManager work together.

Please rebase your patch and give it a commit message of

    Fixes #7821 - start and enable lldpad on all net interfaces

This will link your patch without feature tracking ticket, thanks :-)

jb11211:
screwed up my repo and force rebasing caused this to close. new commit:
https://github.com/theforeman/foreman-discovery-image/pull/66

jb11211:
OK. I screwed up my git repo so I�ve had to redo the PR:

https://github.com/theforeman/foreman-discovery-image/pull/66

This now has the enable-lldpad script called by udev as you suggested and the commit name is as suggested as well.

Hopefully this worked. As I said in the commit I can�t test right now.

PR #30, opened 2015-10-05
Fixes #12052 - status screen correctly displays MAC/IP

lzap:


root/usr/lib64/ruby/vendor_ruby/discovery.rb
```diff
@@ -157,6 +157,14 @@ def env_append(env,string)
   end
 end
 
+def get_mac(interface = 'primary')
+  `nmcli -t -f 802-3-ethernet.mac-address con show #{interface}`.scan(/\w{2}:\w{2}:\w{2}:\w{2}:\w{2}:\w{2}\n/).first.strip rescue 'N/A'
+end
+
+def get_ipv4(interface = 'primary')
+  `nmcli -t -f IP4.ADDRESS con show #{interface}`.scan(/\d+\.\d+\.\d+\.\d+\/\d+\n/).first.strip rescue 'N/A'
```
stbenjam: Is it intentional to show the subnet too?  `192.168.33.131/24 `

lzap: Yeah, I noticed. Why not, when nmcli returns it. The screen is also faster now, no need to call Facter.

stbenjam:
Merged as fec8d807b7b940b68df33bc33a5cd6140e52dec6

PR #31, opened 2015-10-05
Fixes #12007 - status screen fixed for kexec workflow

lzap:
It was not set to server when PXE-less was used (kernel command line is empty
in this case).


stbenjam:
Merged as 3595564b47b8def7961abf09d998e771dc85b8eb

PR #32, opened 2015-10-06
Fixes #12007 - status screen fixed for kexec workflow

lzap:
Simple patch for PXE-less.


lzap:
@stbenjam context is - PXE-less workflow cannot read kernel command line for foreman_url because it is simply not there, passing it along now.

lzap:
Uh I filed this twice! Sry.

PR #33, opened 2015-10-06
Fixes #12069 - pxip/gw no longer required in unattended DHCP mode

lzap:
In Unattended DHCP mode, we were insisting on IP/GW but it should also work
with DHCP. @stbenjam thanks!


stbenjam:
Thanks! Looks good, merged as a75e0efcc627323af42f8ff7cda22649e2de97f4

PR #35, opened 2015-10-08
Refs #12003 - legacy Facts API no longer required

lzap:
@stbenjam This is the last series of patches. Since fdi 3.0+ now listens on
HTTPS by default, we no longer can leverage the standard Smart Proxy Facts API
which requires SSL client auth.

This series (plugin, image, proxy-proxy, proxy-image) adds new API which I
named "inventory_api" not to confuse with original "facts_api".

This is the last one, related:

https://github.com/theforeman/smart_proxy_discovery/pull/9
https://github.com/theforeman/foreman_discovery/pull/223
https://github.com/theforeman/smart_proxy_discovery_image/pull/3

Many thanks for your effort!

### DIFF
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -78,11 +78,6 @@ cat >/etc/foreman-proxy/settings.d/discovery_image.yml <<'CFG'
 :enabled: true
 CFG
 
-cat >/etc/foreman-proxy/settings.d/facts.yml <<'CFG'
----
-:enabled: true
-CFG
-
 cat >/etc/foreman-proxy/settings.d/bmc.yml <<'CFG'
 ---
 :enabled: true


stbenjam:
Hmm, could we just see if smart-proxy would accept a patch to make the facts API authorize_with_ssl_client configurable with a setting? It's a lot of duplicate code.

https://github.com/theforeman/smart-proxy/blob/develop/modules/facts/facts_api.rb#L4

lzap:
Hmm dangerous one, not sure I like it. @domcleal @GregSutcliffe ?

My ultimate goal is to get rid of standard smart proxy plugins and have only node API plugin installed on image.

lzap:
The Facts API was created in this this ticket and I cant say where is it used in Foreman. http://projects.theforeman.org/issues/1797

lzap:
On standard installation this plugin is turned off. I think we can get rid of it completely and move it into separate one perhaps?

domcleal:
> The Facts API was created in this this ticket and I cant say where is it used in Foreman. http://projects.theforeman.org/issues/1797

Foreman doesn't use it, perhaps some people do, but I'd say feel free to propose a deprecation and eventual removal.

lzap:
Yeah it looks so, thanks.

```
commit 90cc95ac2ffba553276af362cff41711ce977529
Refs:
Author:     Roberto <roberto_moral@cable.comcast.com>
AuthorDate: Tue Jul 31 10:51:03 2012 -0400
Commit:     Ohad Levy <ohadlevy@gmail.com>
CommitDate: Tue Jul 31 22:18:45 2012 +0300
```

I will ask on the list and if there are still users, I can refactor to external one. Otherwise I will remove. What you think @stbenjam ?

stbenjam:
Did you intend to move the facts api to a smart_proxy_facts plugin then? Or remove it from the smart proxy and use this inventory api in discovery?

lzap:
Yeah remove and to use inventory_api instead. The idea behind not just moving is to have one single namespace for all image-related APIs.

lzap:
Ping @stbenjam - I'd like to release 4.1 this week with RC1. If there is any chance of merging the last three issues: http://projects.theforeman.org/issues/10294

stbenjam:
Merged as 08245239065a1b2db88cef8d37ec65b61a2fc340, that's the last one for new inventory API, thanks @lzap

lzap:
Thank you so much!

PR #38, opened 2015-10-22
Remaster script works on old images as well

lzap:
For those who want to experiment with remastered 2.X images. Please note after
the host REBOOTs it wont be able to boot the installer. Old images do not
support kexecing yet. Also adds overwrite option for dev purposes.

### DIFF
--- a/aux/remaster/discovery-remaster
+++ b/aux/remaster/discovery-remaster
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [ -z "$2" ]; then
-  echo "Usage: $0 fdi-bootable-x.y.z.iso 'proxy.url=https://192.168.9.1:443 proxy.type=foreman fdi.xyz=abc ...'"
+  echo "Usage: $0 fdi-bootable-x.y.z.iso 'proxy.url=https://192.168.9.1:443 proxy.type=foreman fdi.xyz=abc...' [output.iso]"
   exit 2
 fi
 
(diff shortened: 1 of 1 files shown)


lzap:
And this one @stbenjam is nice to have (since I remaster quite often the new option is *very* handy). Thanks!

stbenjam:
How does this work? The published 2.0 images aren't ISO's, right? Was just trying to test it out.

lzap:
> How does this work? The published 2.0 images aren't ISO's, right? Was just trying to test it out.

Yes, a colleague of us asked me if he can remaster an old build, so I
made a changes. It will boot and register, but kexecing won't work and
since there is no TUI, it's very limited use.

This patch also adds useful extra cmd option :-)

-- 
Later,
 Lukas #lzap Zapletal

stbenjam:
But how do I build an ISO for 2.0? It generates a tar and an img.

lzap:
> But how do I build an ISO for 2.0? It generates a tar and an img.

Yeah, you can either extract the ISO from downstream RPM, or modify the
build script not to delete the resulting ISO file after initrd
extraction.

You don't need to test this, if it works for you with the current image
it's fine to merge this "as is".

-- 
Later,
 Lukas #lzap Zapletal

stbenjam:
I just prefer to test the ones I can, anyway merged as d7792746ff0b9f8d17c89f6e6fe392973955405e, thanks

PR #39, opened 2015-10-26
Fixes #12310 - clear Facter cache on every discovery-register iteration

balexx:


lzap:
Hello,

we have that feature already via `fdi.cachefacts`. By default it is set to 5 in the 3.0.1-3.0.3 images, but it brings race conditions, so we will turn this off to old behavior (you can still override this via the option).

https://github.com/theforeman/foreman-discovery-image/pull/41

I am closing this PR.

lzap:
Few years later I realized the issue, thanks :-)

I created a new PR for the same problem, I misinterpreted the bug. https://github.com/theforeman/foreman-discovery-image/pull/115

PR #40, opened 2015-10-27
Fixes #12322 - disabled resend button for PXE-less

lzap:
Resend button restarts discovery-register service, but in PXE-less environment
this service is inactive. For technical reasons, we just emit message for user
to start over.

### DIFF
--- a/root/usr/lib64/ruby/vendor_ruby/discovery/screen/network.rb
+++ b/root/usr/lib64/ruby/vendor_ruby/discovery/screen/network.rb
@@ -30,7 +30,7 @@ def screen_network mac, ip = cmdline('fdi.pxip', ''), gw = cmdline('fdi.pxgw', '
       return [:screen_network, mac, ip, gw, dns]
     end
     action = Proc.new { configure_network true, mac, ip, gw, dns }
-    [:screen_info, action, "Configuring network via DHCP. This operation can take several minutes to complete.", "Unable to bring network via DHCP",
+    [:screen_info, action, "Configuring network. This operation can take several minutes to complete.", "Unable to bring up network",
       [:screen_foreman, mac, gw],
       [:screen_network, mac, ip, gw, dns]]
   else
--- a/root/usr/lib64/ruby/vendor_ruby/discovery/screen/status.rb
+++ b/root/usr/lib64/ruby/vendor_ruby/discovery/screen/status.rb
@@ -68,8 +68,12 @@ def screen_status status = generate_info, active_button = 0
   elsif answer == b_ssh
     :screen_ssh
   elsif answer == b_resend
-    command("rm -f /tmp/discovery-http*")
-    command("systemctl reload discovery-register")
+    if cmdline('BOOTIF')
+      command("rm -f /tmp/discovery-http*")
+      command("systemctl reload discovery-register")
+    else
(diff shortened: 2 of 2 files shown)


root/usr/lib64/ruby/vendor_ruby/discovery/screen/network.rb
```diff
@@ -30,7 +30,7 @@ def screen_network mac, ip = cmdline('fdi.pxip', ''), gw = cmdline('fdi.pxgw', '
       return [:screen_network, mac, ip, gw, dns]
     end
     action = Proc.new { configure_network true, mac, ip, gw, dns }
-    [:screen_info, action, "Configuring network via DHCP. This operation can take several minutes to complete.", "Unable to bring network via DHCP",
+    [:screen_info, action, "Configuring network. This operation can take several minutes to complete.", "Unable to bring network via DHCP",
```
lzap: Reworded, this screen is also shown for *static* configuration...

root/usr/lib64/ruby/vendor_ruby/discovery/screen/status.rb
```diff
@@ -68,8 +68,12 @@ def screen_status status = generate_info, active_button = 0
...
     :screen_ssh
   elsif answer == b_resend
-    command("rm -f /tmp/discovery-http*")
-    command("systemctl reload discovery-register")
+    if cmdline('BOOTIF')
+      command("rm -f /tmp/discovery-http*")
+      command("systemctl reload discovery-register")
+    else
+      Newt::Screen.win_message("Not supported", "OK", "Resending not possible in PXE-less, reboot and start over.")
+    end
```
lzap: For technical reasons, we just error out in case user presses this button, I do not want to re-implement whole workflow to be able to resend the facts from GUI now (I don't have parameters which are being passed among "screen_" methods).

root/usr/lib64/ruby/vendor_ruby/discovery/screen/network.rb
stbenjam: What about the second error message?  "Unable to bring network via DHCP" - could you hit that in static network config?

lzap: Good catch, fixing!

root/usr/lib64/ruby/vendor_ruby/discovery/screen/network.rb
```diff
@@ -30,7 +30,7 @@ def screen_network mac, ip = cmdline('fdi.pxip', ''), gw = cmdline('fdi.pxgw', '
       return [:screen_network, mac, ip, gw, dns]
     end
     action = Proc.new { configure_network true, mac, ip, gw, dns }
-    [:screen_info, action, "Configuring network via DHCP. This operation can take several minutes to complete.", "Unable to bring network via DHCP",
+    [:screen_info, action, "Configuring network. This operation can take several minutes to complete.", "Unable to bring network",
```
stbenjam: Sorry, nitpick - "bring up network" is probably better

stbenjam:
Merged as 4d3c341c14af2f4b03702f5472df2e3660e22624, thanks @lzap !

PR #41, opened 2015-10-27
Fixes #12323 - def number of register calls is zero

lzap:
If network is down (typical problem on slower DHCP networks) during initial discovery, we should still allow users to override this value. But it should not be 5 by default to prevent race conditions described above.


lzap:
@stbenjam was hoping for two quick merges, cosmetic issues. Here and https://github.com/theforeman/foreman-discovery-image/pull/40.

lzap:
This one @stbenjam

stbenjam:
Merged as 1a16ca7bfa64395f180df513b18228c1be4294ca

PR #42, opened 2015-10-29
Fixes #12275 - fixed error messages and NIC sort

lzap:
Here is one more @stbenjam. Unfortunately my last patch `.sort` broke master,
forgot to add `.each`. This fixes it and also improves our logging and error
states. Also STDOUT is silenced for facter as there is no other way to tell
facter not to raise errors when no IP is present on the system.


lzap:
This one @stbenjam

stbenjam:
Merged as 8347043517aec9a4b621b529ee5f0a18dd8ba71e, thanks @lzap!

PR #43, opened 2015-11-02
Fixes #12340 - resend button shows countdown now

lzap:
This one @stbenjam as well. When using Resend button, we were just restarting
discovery-register service immediately returning to status screen (which showed
out-of-date info). Users were asked to hit Status screen to redraw which was
unintuitive.

Now, we switch over to countdown screen and after 10 seconds (or if we already
know the result via temp files) we end up on Status screen with correct info.


lzap:
When merging this, please merge docs as well thanks: https://github.com/theforeman/theforeman.org/pull/457

lzap:
Damn, I will rebase now.

lzap:
I need to file new PR, bad base branch.

PR #47, opened 2015-11-09
Fixes #12420 - primary detected correctly for PXE-less

lzap:
The root cause is our discovery_bootif fact which only accepts BOOTIF option,
but we should also use fdi.pxmac when BOOTIF is not present. Also added one
more fallback if none present - tries to detect first IFACE with link. And if
that fails, we fallback to facter's "macaddress" which is pretty less "random"
and generally a bad choice.

### DIFF
--- a/root/usr/lib64/ruby/vendor_ruby/discovery.rb
+++ b/root/usr/lib64/ruby/vendor_ruby/discovery.rb
@@ -68,6 +68,23 @@ def cmdline option=nil, default=nil
   end
 end
 
+def detect_first_nic_with_link
+  log_debug "Detecting the first NICs with link"
+  mac = ''
+  Dir.glob('/sys/class/net/*').sort.each do |ifn|
+    name = File.basename ifn
+    next if name == "lo"
+    mac = File.read("#{ifn}/address").chomp rescue ''
+    link = File.read("#{ifn}/carrier").chomp == "1" rescue false
+    if link
+      log_debug "Interface with link found: #{mac} (#{name})"
+      break
+    end
+  end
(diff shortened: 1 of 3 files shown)


root/usr/lib64/ruby/vendor_ruby/discovery.rb
```diff
@@ -68,6 +68,22 @@ def cmdline option=nil, default=nil
   end
 end
 
+def detect_first_nic_with_link(default = nil)
+  log_debug "Detecting the first NICs with link"
+  mac = default
```
stbenjam: I don't think this works as the last fallback because for every interface in the Dir block, you're changing the value of `mac`

lzap: > I don't think this works as the last fallback because for every interface in the Dir block, you're changing the value of `mac`

Good catch, the method was implemented in a way that it returns at least
some interface if there are none with link. I am removing the default
parameter and removing the last fallback to use "macaddress" fact which
proved to be very unstable source of information. It changes over time,
reboot or even when you add new hardware via hotplug.

Therefore the link identification is the last try to get the primary
interface.

-- 
Later,
 Lukas #lzap Zapletal

lzap:
RPM errata blocker, thanks!

lzap:
@stbenjam pls let me know on IRC when merging this, I will immediately tag new upstream release and rebase downstream.

stbenjam:
Merged as 47d3d43da11e55244a85f56085140255045397be, thanks @lzap

PR #48, opened 2015-11-10
Fixes #12429 - foreman-proxy waits till network is fully prepared

kmadac:
Pull request hooks foreman-proxy to NetworkManager-wait-online.service instead of network.target and increase the timeout to 120s of nm-online to wait long enough on machines when we need to wait till NM configures lot of NICs (10+).


lzap:
Ahoj Kamile, good catch yeah indeed, I haven't realized that proxy now actually requires the primary connection. There is one drawback tho, discovery supports PXE-less environment when network is not being initialized at all during boot. In this case the service fails to start due to this error which is expected and it is started afterwards manually from the GUI.

To your patch, you should never use the `NetworkManager-wait-online.service` directly, use `network-online.target` instead. Also the extra `nm-online` command is not required I guess. Let me try to fix this in a different way.

lzap:
Bringing a different solution, a more complex one.

PR #49, opened 2015-11-10
Fixes #12430 - unattended mode sends custom facts now

lzap:
Sachin @sghai found an issue when using unattended, the custom facts are *not*
sent at all. Mistyped parenthesis there. Thanks @stbenjam!


stbenjam:
Merged as bfa733c7a12092af8251c09d66e1adf25cd90fad, thanks!

PR #50, opened 2015-11-11
Fixes #12429 - fixed network race conditions

lzap:
This patch replaces
https://github.com/theforeman/foreman-discovery-image/pull/48

Makes sure foreman-proxy starts after network-online. Also fixes other possible
related race conditions. Comments inline.

### DIFF
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -48,7 +48,10 @@ systemctl disable ipmi.service
 echo " * open foreman-proxy port via firewalld"
 firewall-offline-cmd --zone=public --add-port=8443/tcp --add-port=8448/tcp
 
-echo " * setting up foreman proxy"
+echo " * setting up foreman proxy service"
+sed -i 's/After=.*/After=basic.target network-online.target nm-prepare.service/' /usr/lib/systemd/system/foreman-proxy.service
+sed -i 's/Wants=.*/Wants=basic.target network-online.target nm-prepare.service/' /usr/lib/systemd/system/foreman-proxy.service
+sed -i '/\[Unit\]/a ConditionPathExists=/etc/NetworkManager/system-connections/primary' /usr/lib/systemd/system/foreman-proxy.service
 sed -i '/\[Service\]/a EnvironmentFile=-/etc/default/discovery' /usr/lib/systemd/system/foreman-proxy.service
 sed -i '/\[Service\]/a ExecStartPre=/usr/bin/generate-proxy-cert' /usr/lib/systemd/system/foreman-proxy.service
 sed -i '/\[Service\]/a PermissionsStartOnly=true' /usr/lib/systemd/system/foreman-proxy.service
--- a/root/etc/systemd/system/discovery-fetch-extensions.service
+++ b/root/etc/systemd/system/discovery-fetch-extensions.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=Download zip extensions to this image
-Wants=discovery-start-extensions.service
+Wants=network-online.target discovery-start-extensions.service
 Before=discovery-start-extensions.service
+After=network-online.target
 
 [Service]
 Type=oneshot
--- a/root/etc/systemd/system/discovery-menu.service
+++ b/root/etc/systemd/system/discovery-menu.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Display interactive TUI on tty1
 Wants=basic.target
-After=basic.target network.target network-online.target nss-lookup.target
+After=basic.target network.target nss-lookup.target
 ConditionPathExists=/dev/tty1
 
 [Service]
--- a/root/etc/systemd/system/discovery-register.service
+++ b/root/etc/systemd/system/discovery-register.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Register this host in Foreman
-After=basic.target network.target network-online.target nss-lookup.target foreman-proxy.service discovery-register.service
+Wants=network-online.target
+After=basic.target network-online.target nss-lookup.target foreman-proxy.service
 
 [Service]
 Type=simple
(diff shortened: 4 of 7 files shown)


lzap:
@stbenjam @kmadac here is a scratchbuild for test: http://downloads.theforeman.org/discovery/lzap/scratch/

22-discovery.ks
```diff
@@ -48,7 +48,10 @@ systemctl disable ipmi.service
 echo " * open foreman-proxy port via firewalld"
 firewall-offline-cmd --zone=public --add-port=8443/tcp --add-port=8448/tcp
 
-echo " * setting up foreman proxy"
+echo " * setting up foreman proxy service"
+sed -i 's/After=.*/After=basic.target network-online.target nm-prepare.service/' /usr/lib/systemd/system/foreman-proxy.service
+sed -i 's/Wants=.*/Wants=basic.target network-online.target nm-prepare.service/' /usr/lib/systemd/system/foreman-proxy.service
+sed -i '/\[Unit\]/a ConditionPathExists=/etc/NetworkManager/system-connections/primary' /usr/lib/systemd/system/foreman-proxy.service
```
lzap: Makes sure this starts really when Network Online target reaches "started" state. Also skips this service when doing PXE-less boot (primary connection is not configured in that case).

root/etc/systemd/system/discovery-fetch-extensions.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=Download zip extensions to this image
-Wants=discovery-start-extensions.service
+Wants=network-online.target discovery-start-extensions.service
 Before=discovery-start-extensions.service
+After=network-online.target
```
lzap: Extensions should also be started after network is fully online.

root/etc/systemd/system/discovery-menu.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Display interactive TUI on tty1
 Wants=basic.target
-After=basic.target network.target network-online.target nss-lookup.target
+After=basic.target network.target nss-lookup.target nm-prepare.service
```
lzap: On the other hand, menu should NOT require network online, because in PXE-less network is usually not configured at all in this workflow.

root/etc/systemd/system/discovery-register.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Register this host in Foreman
-After=basic.target network.target network-online.target nss-lookup.target foreman-proxy.service discovery-register.service
+Wants=network-online.target
+After=basic.target network-online.target nss-lookup.target foreman-proxy.service discovery-register.service nm-prepare.service
```
lzap: Ditto above.

root/etc/systemd/system/nm-prepare.service
```diff
@@ -2,11 +2,13 @@
 Description=Prepares NetworkManager configuration for primary interface
 Wants=basic.target
 Before=NetworkManager.service
+ConditionKernelCommandLine=BOOTIF
```
lzap: Until now, we have been failing with exit code 1 when ISO was booted without BOOTIF effectively not configuring the primary interface. This was leaving an error in journal, this way it's prevented completely.

root/usr/bin/generate-proxy-cert
```diff
@@ -3,7 +3,7 @@
 # When booted from ISO, foreman-proxy starts up by default but since
 # network is not yet configured, we cannot continue. This makes
 # the service to fail.
-[ -f /etc/NetworkManager/system-connections/primary ] || exit 1
+[ -f /etc/NetworkManager/system-connections/primary ] || exit 101
```
lzap: More transparent when this fails for this reason, 1 is too generic number.

root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
```diff
@@ -84,6 +84,7 @@ def configure_network static, mac, ip=nil, gw=nil, dns=nil
   command("nmcli connection reload")
   command("nmcli connection down primary")
   result = command("nmcli connection up primary", false)
+  command("nm-online -s -q --timeout=45") unless static
```
lzap: This method is called from TUI on all network configuration actions, this actually makes sure its online when using DHCP, so proxy start on the very next line won't fail.

lzap:
@schiermi might be interested in this patch as well.

lzap:
Ignore the incorrect version in the filename, should be it: http://downloads.theforeman.org/discovery/lzap/scratch/fdi-bootable-2.1.1.iso

lzap:
I have just learned about non-forking services and `systemd-notify` command to change status to "started". I will redo this patch with this in mind. I think this will fix all the problems we have with systemd/nm!

http://www.freedesktop.org/software/systemd/man/systemd.service.html (see Type)
http://www.freedesktop.org/software/systemd/man/systemd-notify.html

lzap:
@stbenjam looks like another rebase downstream, if you can take a look on this one it is a bonus!

lzap:
btw I have checked and we use "oneshot" type for all the services that needs some ordering, which means they are considered "started" once the process finishes.

22-discovery.ks
kmadac: +1 - I just tested this part of code in our environment and it works (Dell PowerEdge R730, 9 NICs in each) - discovery image registers servers via foreman-proxy and foreman reboots and deploys servers automatically.

lzap:
Ok we have a report this works fine, thank you!

lzap:
@stbenjam I plan one more minor upstream release before Christmas, if you can find some time to merge this one. Annoying bug, hopefully the final resolution to this kind of problems. Thanks!

stbenjam:
Do we still need this one?

lzap:
Absolutely, @stbenjam! Just make sure it boots up via PXE and ISO.

stbenjam:
Tested and works fine, merged as 0c18ba2a6d04e5105db1e2085fe69f091b6922c7. Thanks @lzap!

PR #54, opened 2015-11-30
RHBZ#1285810 - RHEL/CentOS 7.2 dracut dependency fixed

lzap:
Due to missing dependency in dracut package, the image fails to build with RHEL
7.2 base. This prevents from missing the dependency.

Symptoms: Init ram disk is missing from the ISO file.

### DIFF
--- a/20-packages.ks
+++ b/20-packages.ks
@@ -15,6 +15,13 @@ tftp
 lldpad
 isomd5sum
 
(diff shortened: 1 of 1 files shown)


stbenjam:
Merged as c45c690d3160b736725d9746a3a6209b5d5b0500

PR #55, opened 2015-11-30
RHBZ#1285810 - RHEL/CentOS 7.2 dracut dependency fixed (dd)

lzap:
I need one additional change, it builds upstream but brew does not support
provides perhaps, the correct package name is "coreutils" really.

### DIFF
--- a/20-packages.ks
+++ b/20-packages.ks
@@ -19,7 +19,7 @@ isomd5sum
 dracut
 tar
 gzip
-dd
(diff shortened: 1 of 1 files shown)


stbenjam:
Hmm, yea, I'd have thought it would've pulled it in from provides.

Anyway, merged 2bcb4fc1219bb222d9f2ec8b5e4b5a10a09eb077

PR #56, opened 2015-11-30
RHBZ#1285810 - RHEL/CentOS 7.2 dracut dependency fixed (tar)

lzap:
@stbenjam shame on me, this was always working upstream because we do not yet
build against CentOS 7.2 but downstream this was a problem. This is now the
final fix.

### DIFF
--- a/20-packages.ks
+++ b/20-packages.ks
@@ -15,12 +15,9 @@ tftp
 lldpad
 isomd5sum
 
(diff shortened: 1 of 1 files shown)


stbenjam:
26cd5826aab74206fa2c09eae4ccd823254830d8

lzap:
Thanks!

PR #57, opened 2015-11-30
A pxeconfig option to specify root's password in encrypted format

korekhov:


lzap:
Nice, thanks. Can you maybe determine if fdi.rootpw starts with $n$ and then use the appropriate option? This way it will work in both cases.

22-discovery.ks
```diff
@@ -101,7 +101,7 @@ echo "ForwardToConsole=no" >> /etc/systemd/journald.conf
 systemctl enable journalctl.service
 
 echo " * configuring foreman-proxy"
-sed -i 's|.*:log_file:.*|:log_file: SYSLOG|' /etc/foreman-proxy/settings.yml
+##sed -i 's|.*:log_file:.*|:log_file: SYSLOG|' /etc/foreman-proxy/settings.yml
```
lzap: Please remove this there is another PR pending to fix this.

root/etc/rc.d/rc.local
```diff
@@ -8,6 +8,11 @@ if [[ ! -z "$KCL_FDI_ROOTPW" ]]; then
   echo "root:$KCL_FDI_ROOTPW" | chpasswd >/dev/null
 fi
 
+# set encrypted root password - overrides non-encrypted password
+if [[ ! -z "$KCL_FDI_ENCRYPTED_ROOTPW" ]]; then
+  echo "root:$KCL_FDI_ENCRYPTED_ROOTPW" | sed -e 's/"//g' -e "s/'//g" | chpasswd -e >/dev/null
```
lzap: Good idea to remove the quotes, can you do it in one regexp?

korekhov:
Updated this PR with requested changes.

root/etc/rc.d/rc.local
```diff
@@ -5,7 +5,14 @@ exportKCL
 
 # setup root password
 if [[ ! -z "$KCL_FDI_ROOTPW" ]]; then
-  echo "root:$KCL_FDI_ROOTPW" | chpasswd >/dev/null
+  # remove possible double and/or single quotes surrounding password string
+  KCL_FDI_ROOTPW=`echo $KCL_FDI_ROOTPW | sed "s/\('\|\"\)//g"`
```
lzap: AFAIK encrypted passwords are safe, no need to quote them at all, please remove. Explanation here: http://man7.org/linux/man-pages/man3/crypt.3.html#NOTES

root/etc/rc.d/rc.local
```diff
@@ -5,7 +5,14 @@ exportKCL
...
+  # if password starts with dollar sign, set an encrypted flag
+  echo $KCL_FDI_ROOTPW | grep '^\$.*$' > /dev/null
```
lzap: We are using GNU Bash here, test with internal:

http://stackoverflow.com/questions/2172352/in-bash-how-can-i-check-if-a-string-begins-with-some-value

root/etc/rc.d/rc.local
korekhov: This is to increase foolproof of this code so it it does not break even if a user puts the password in quotes, so the following forms of passwords in pxeconfig will just work w/o issues:
fdi.rootpw="$1$_redhat_$i3.3Eg7ko/Peu/7Q/1.wJ/"
fdi.rootpw='$1$_redhat_$i3.3Eg7ko/Peu/7Q/1.wJ/'
fdi.rootpw=$1$_redhat_$i3.3Eg7ko/Peu/7Q/1.wJ/
fdi.rootpw=redhat

What is your concern having this here?

root/etc/rc.d/rc.local
korekhov: Sure, I can change that. The way it was done before would work not only in specific versions of shell, but that's fine.

root/etc/rc.d/rc.local
lzap: Well, what you do is you strip all quotes from the string. Some people use these characters in passwords, when entered in plain text you break the feature. I understand the motivation, you can do it either if its a starting and ending quote (and only in case both are the same), or simply don't do it. I am for the latter.

root/etc/rc.d/rc.local
lzap: > Sure, I can change that. The way it was done before would work not only in specific versions of shell, but that's fine.

Yeah, we are safe here, GNU Bash only.

-- 
Later,
 Lukas #lzap Zapletal

lzap:
Please don't get me wrong, I'd prefer if you delete the quotes magic line from the patch. It's a great addition, we are good to go. Please squash into one single commit.

root/etc/rc.d/rc.local
```diff
@@ -5,7 +5,13 @@ exportKCL
 
 # setup root password
 if [[ ! -z "$KCL_FDI_ROOTPW" ]]; then
-  echo "root:$KCL_FDI_ROOTPW" | chpasswd >/dev/null
+  # remove possible double and/or single quotes surrounding password string
+  KCL_FDI_ROOTPW=`echo $KCL_FDI_ROOTPW | sed -e "s/^'\(.*\)'$/\1/" -e 's/^"\(.*\)"$/\1/'`
```
lzap: Ping, can you remove this line and squash please?

korekhov: But this only removes possible *surrounding* single or double-quotes and only in pairs. Isn't that want you wanted?
If we don't remove them, next check will fail when the user puts the quotes around.

As for squash - how do I do that? Not a git power-user, sorry...

dLobatog: `git rebase -i HEAD~numberofcommits` , then you'll be prompted and you can squash all your commits by changing all of the 'pick' but the first one to 'squash'

korekhov:
Will resubmit a PR after squashing.

korekhov:
https://github.com/theforeman/foreman-discovery-image/pull/63

lzap:
> Will resubmit a PR after squashing.

For the record, force-pushing is okay on github. It's just Linus
Torvalds who hates it :-)

-- 
Later,
 Lukas #lzap Zapletal

PR #58, opened 2015-12-01
Fixes #12641 - Refresh facts errors and adds new entry mac000000000000

lzap:
47d3d43da11e introduced fallback via detect_first_nic_with_link method, but
since this was called within Proxy context, SYSLOG was raising an exception
"already opened socket". Therefore we need to detect the context in `log_`
methods and log via native Proxy logging library instead.

This patch fixes this problem and also improves logging and detection code.


stbenjam:
Merged as 1f884b729ebf84308cd18d5c44cf7bf7d57a1ded

PR #51, opened 2015-11-13
Cleaned up foreman-proxy configuration

lzap:
A bit messy after all the changes, this cleans things up.

PR #60, opened 2015-12-10
when building image on CentOS, option log_file: STDOUT  conflicts wit…

styshoo:
when building image on CentOS, option log_file: STDOUT  conflicts with option daemon: true.


lzap:
This patch solves this, feel free to give it :+1: if you test it.

https://github.com/theforeman/foreman-discovery-image/pull/51/files

PR #62, opened 2016-02-03
F21

michaelrice:
When trying to build the image you would get a failure caused by a 404

    Error creating Live CD : failure: repodata/repomd.xml from foreman-plugins-f19: [Errno 256] No more mirrors to try.
    http://yum.theforeman.org/plugins/nightly/f19/x86_64/repodata/repomd.xml: [Errno 14] HTTP Error 404 - Not Found

Updated links and renamed files for f21 since that is the newest repo I could find.


stbenjam:
Thanks, can you squash to 1 commit?

michaelrice:
Done!

dLobatog:
Merged as 85880b7c0acac9301c8893c02fa830949da45d71, thanks @michaelrice !

lzap:
Thanks Michael! I haven't tested on F21 yet, hows the image. Does it work as expected?

michaelrice:
@lzap I just use the base image in another project, and it works, but I do not use it as it is meant to be used.. like I do not have foreman at all.. so Im not sure

PR #63, opened 2016-02-11
root password can be specified in encrypted format

korekhov:


dLobatog:
@korekhov Just for the record, you can just `git push -f` after you squash to the same branch, no need to submit 2 PRs. Anyhow, merging!

dLobatog:
Merged as c4750cdda5fbfa68346c3467fae52fdaa641c31a, thanks @korekhov!

README.md
```diff
@@ -163,6 +163,15 @@ ssh and set root password using the following kernel command line options:
...
 ```
 
+Root password can also be specified in encrypted form (using 'redhat' as
+an example below). Single and/or double quotes around password are 
+recommended to be used to prevent possible special characters
+interpretation.
+
+```
+fdi.rootpw='$1$_redhat_$i3.3Eg7ko/Peu/7Q/1.wJ/'
+```
```
lzap: I still don't like the quotes, kernel command line does not have shell-like parsing or interpolation. More than that, passwd library does use base64 encoding which is safe.

PR #46, opened 2015-11-05
Drop unwanted images to save 13 MB

lzap:
Plus there is additional patch in livecd-tools to save additional 33 MB!

https://github.com/rhinstaller/livecd-tools/pull/18

Our BMI is back to normal now ;-)

PR #64, opened 2016-02-24
Improved discovery-debug script

lzap:
Introduced in https://github.com/theforeman/foreman-discovery-image/pull/46
sorry!

### DIFF
--- a/25-minimize.ks
+++ b/25-minimize.ks
@@ -63,7 +63,7 @@ echo " * locking root account"
 passwd -l root
 
 echo " * store list of packages sorted by size"
-rpm -qa --queryformat '%{SIZE} %{NAME}\n' | sort -n -r > /usr/PACKAGES-LIST
+rpm -qa --queryformat '%{SIZE} %{NAME}%{VERSION}%{RELEASE}\n' | sort -n -r > /usr/PACKAGES-LIST
 
 echo " * cleaning up yum cache and removing rpm database"
 yum clean all
--- a/root/usr/bin/discovery-debug
+++ b/root/usr/bin/discovery-debug
@@ -2,22 +2,17 @@
 
 source /etc/default/discovery
 
-if [ "$1" != "--tui" ]; then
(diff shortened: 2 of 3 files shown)


stbenjam:
Didn't the old one include the package version as well, is there a reason we're not doing that anymore?

lzap:
Sorry I am not following...

lzap:
@stbenjam can you elaborate?

I also simplified the debug script in this patch, the `--tui` option was I think unnecessary (it's purpose was to remove some things when called from TUI). I also removed one extra button - Network. It's always been already included in System log. I am making some space for another buttons for the future.

stbenjam:
PACKAGE-LIST is just the package names.

lzap:
Oh yeah, well since this is read only image, versions are not relevant in this context - RPM database is deleted, no chance to upgrade a package.

stbenjam:
Right, but I wanted to see the version of the smart-proxy RPM to make sure what nightly it had installed. I suppose I could've looked at the build-live-cd output but if we're including a package list, why not have the version?

lzap:
> Right, but I wanted to see the version of the smart-proxy RPM to make sure what nightly it had installed. I suppose I could've looked at the build-live-cd output but if we're including a package list, why not have the version?

Sure, updated:

```
rpm -qa --queryformat '%{SIZE} %{NAME}%{VERSION}%{RELEASE}\n' | head -n5
432073 perl-Math-BigInt1.9997350.fc23
62711 lockdev1.0.40.20.20111007git.fc23
245772 mathjax-sansserif-fonts2.4.02.fc23
2295691 jomolhari-fonts0.00320.fc23
20911 python3-pwquality1.3.02.fc23
```

-- 
Later,
 Lukas #lzap Zapletal

stbenjam:
Thanks!! Merged as ad8c2b7ed74a3f2b3b3c6cdd2951ab20ac6b12df

PR #65, opened 2016-02-29
Fixes #8540 - hostname is no longer sent in DHCP request

lzap:
We currently send "fdi" hostname in a DHCP request. This can create secondary
lease when PXE DHCP lease was sent with a different name (which is likely).
This saves one extra lease in case BIOS/EFI also sends without any hostname
which is the case for most of them.

To test this: Boot a host without this patch, you should see "fdi" in the
leases file. Reboot with a different IP address and this patch, no hostname
should be sent.

### DIFF
--- a/root/usr/bin/nm-configure
+++ b/root/usr/bin/nm-configure
@@ -57,6 +57,7 @@ autoconnect-priority=1
 mac-address=$mac
 [ipv4]
 method=auto
+dhcp-send-hostname=false
 dhcp-timeout=$timeout
 [ipv6]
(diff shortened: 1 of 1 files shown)


lzap:
@dLobatog also quick and easy merge - context - we consume extra DHCP lease which is unnecessary. Cosmetic bug but nice to have in the next release.

dLobatog:
Merged as 92aaa220d37c7b232e3a66e9bb90d4c5696d9123, thanks @lzap!

PR #66, opened 2016-02-29
Fixes #7821 - start and enable lldpad on all net interfaces

jb11211:
This change enables the lldpad daemon, turns on lldp for
all interfaces (excluding lo) and then uses the openlldp.rb
written by razorsedge to report the lldp data in facter

messed up the old PR so I'm cutting another one. 

It's been months since I've had a working foreman-discover-image test env. so this is untested.

### DIFF
--- a/root/usr/bin/enable-lldp
+++ b/root/usr/bin/enable-lldp
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+enable_lldp_int() {
+  local int=$1
+  shift
+  local ip=$1
+  shift
+
+  variables="$@"
+
+  lldptool -L -i $int adminstatus=rxtx
+
+  for var in $variables; do
+    lldptool -i $int -T -V $var enableTx=yes
+  done
+
+}
+
+for inter in $(ls /sys/class/net/ | grep -v "lo"); do
+ enable_lldp_int $inter $ipaddress portDesc sysName sysDesc sysCap
+done
+
+exit 0
--- a/root/usr/share/fdi/facts/openlldp.rb
+++ b/root/usr/share/fdi/facts/openlldp.rb
@@ -1,0 +1,116 @@
+# grabbed from:
+# https://raw.githubusercontent.com/razorsedge/puppet-openlldp/master/lib/facter/openlldp.rb
+
+#Copyright (C) 2012 Mike Arnold <mike@razorsedge.org>
+#
+#Licensed under the Apache License, Version 2.0 (the "License");
+#you may not use this file except in compliance with the License.
+#You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


root/usr/bin/enable-lldp
```diff
@@ -0,0 +1,23 @@
...
+
+  for var in $variables; do
+    lldptool -i $int -T -V $var enableTx=yes
+  done
+
+}
+
+for inter in $(ls /sys/class/net/ | grep -v "lo"); do
+ enable_lldp_int $inter $ipaddress portDesc sysName sysDesc sysCap
+done
```
lzap: This command is executed for each individual device as a parameter. Remove the loop here please, you will have the device name as first parameter.

```shell
    dev=$(basename $1)
    [[ "$dev" == "lo" ]] && return
    enable_lldp_int $dev ...
```
Whole script can be simplified a lot. Also please add output (it will go to journald):

```
echo "Configuring LLDP on interface $dev"
```

So we can check the logs if it gets called properly.

root/usr/share/fdi/facts/openlldp.rb
```diff
@@ -0,0 +1,116 @@
...
+    'MTU'          => '0x00120f04',
+  }
+
+  # Remove interfaces that pollute the list (like lo and bond0).
+  Facter.value('interfaces').split(/,/).grep_v(/^lo$|^bond[0-9]/).each do |interface|
+    # Loop through the list of LLDP TLVs that we want to present as facts.
+    lldp.each_pair do |key, value|
+      Facter.add("lldp_neighbor_#{key}_#{interface}") do
+        setcode do
+          result = ""
```
lzap: The result must be `nil` otherwise `facter` returns empty facts and we don't want that.

Other that that, it works. But the loop is useless and it also returns empty facts for interfaces which do not support LLDP.

Please fix the two guys here, rebase and we are finally good to go :-)

I also found a UI bug, LLDP facts were not displayed properly. Filed a bug: http://projects.theforeman.org/issues/14010

jb11211:
I don't use theforeman anymore and cannot devote anymore time to this.

Best of luck with this patch I hope it helps.

beddari:
Thanks for your work :+1:

lzap:
Oh okay, many thanks for this. I will merge and squash the two corrections into your commit, so the authorship remains untouched.

lzap:
Hmmm I haven't realized that the damon must be running. Socket activation does not work with lldpad. I need to take a look tomorrow.

lzap:
Ok this one was tough, socket activation did not help because sockets are not yet initialized when udev rules are triggering. Solved it via udev-activated service that only starts after LLDP daemon:

    ACTION=="add", SUBSYSTEM=="net", NAME!="lo", TAG+="systemd", ENV{SYSTEMD_WANTS}="enable-lldp@%k.service"

Merged as 672443f062b524454374ac6b08b3194645fbe4af thank you!

PR #67, opened 2016-03-01
Drop unwanted UEFI/MacOS loaders from PXE image

lzap:
Although I pushed a change into livecd-tools project to drop unwanted
bootloaders from PXE image, the project seems to be dead now (no release
planned for Fedora 24). The patch I provided saves 33 MB (!) of data from the
PXE image. This patch pulls the script into our repo:

https://github.com/rhinstaller/livecd-tools/commit/8067be50907da9461e442c11a664c89e066ccac6

The project is being replaced with Lorax, new generation of LiveCD building
tools which is currently used for Fedora:

https://lorax.readthedocs.org/en/latest/livemedia-creator.html

We will likely move on to Lorax soon, I don't know if it works with CentOS 7
yet. It will be a big change here, everything seems to be different now. But it
looks like we will be able to boot from HTTP (live: option).

To test this, build from master, then build with this patch and compare the
generated PXE init ram disk size (not ISO file, the ISO remains the same).

### DIFF
--- a/build-livecd-root
+++ b/build-livecd-root
@@ -19,7 +19,7 @@ if [ $(id -u) != 0 ]; then
   exit 1
 fi
 
-which livecd-creator isohybrid livecd-iso-to-pxeboot >/dev/null || ( echo "Command(s) missing, install required tools" && exit 2 )
+which livecd-creator isohybrid >/dev/null || ( echo "Command(s) missing, install required tools" && exit 2 )
 
 trap cleanup EXIT
 
@@ -49,7 +49,7 @@ isohybrid --partok --uefi fdi-bootable.iso
 mv fdi-bootable.iso $srcdir/fdi-bootable-${last_tag#release-}.iso
 
 echo "* Converting to initrd"
-livecd-iso-to-pxeboot $tmpdir/fdi.iso
+$srcdir/aux/livecd-iso-to-pxeboot $tmpdir/fdi.iso
 
 if [ $? -ne 0 ]; then
     KEEP_TMPDIR=yes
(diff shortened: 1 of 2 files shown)


lzap:
Thanks @stbenjam, the current release is quite big for comparison (fresh build from today):

http://downloads.theforeman.org/discovery/releases/3.1/

build-livecd-root
```diff
@@ -49,7 +49,7 @@ isohybrid --partok --uefi fdi-bootable.iso
 mv fdi-bootable.iso $srcdir/fdi-bootable-${last_tag#release-}.iso
 
 echo "* Converting to initrd"
-livecd-iso-to-pxeboot $tmpdir/fdi.iso
+aux/livecd-iso-to-pxeboot $tmpdir/fdi.iso
```
stbenjam: This seems to be executed from within the tmp directory, so it fails for me:

```
+ echo '* Converting to initrd'
* Converting to initrd
+ aux/livecd-iso-to-pxeboot /tmp/build-fdi-lqLbEx/fdi.iso
./build-livecd-root: line 52: aux/livecd-iso-to-pxeboot: No such file or directory
+ '[' 127 -ne 0 ']'
+ KEEP_TMPDIR=yes
```

lzap: Oh I missed it and checked an old kernel when I was testing that. Thanks.

```
-rw-r--r-- 1 lzap lzap 255M Mar  4 09:52 fdi-image-3.1.0.tar
-rw-r--r-- 1 lzap lzap 215M Apr 26 12:50 fdi-image-3.1.1.tar
```

Amended the change.

stbenjam:
Sorry for the delay, image builds and looks good.  Merged as b0708d82ad7526b1abf80fcd4c0e61984ad4ed04

PR #69, opened 2016-03-17
Fixes #14247 - prevents proxy from timing out during startup

lzap:
Setting 30 secs looked like a sane value to me, but it's indeed longer than DNS query timeout which can happen on systems with incorrect DNS setup.

### DIFF
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -95,8 +95,6 @@ cat >/etc/foreman-proxy/settings.d/logs.yml <<'CFG'
 CFG
 
 echo " * setting up systemd"
-echo "DefaultTimeoutStartSec=30s" >> /etc/systemd/system.conf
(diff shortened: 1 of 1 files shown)


lzap:
To give you context - default value is 90 seconds and I shortened this, but it's causing issues with slow DNS queries to some users. Let's leave it on the default option.

lzap:
@dLobatog rebased this is quick merge too - context is that the systemd default (90 secs) is more sane than what I provided myself, that's really it

lzap:
This is easy patch as well, if you will be testing the other one @stbenjam or @dLobatog please merge.

I would like to make new release soon, these are both low hanging fruit:

* https://github.com/theforeman/foreman-discovery-image/pull/65
* https://github.com/theforeman/foreman-discovery-image/pull/76

dLobatog:
I've been a victim of this one, glad you fixed it :smile:

dLobatog:
Thanks @lzap !

PR #70, opened 2016-03-22
Fixes #14315 - fixed proxy logger and enabled log buffer

lzap:
To test this, discover host and refresh its facts, watch journald - it should
contain some logs during refresh facts like primary NIC detection. Messages
like "Detecting the first NICs with link" should be there and no errors about
already opened sylog present.

This patch also enables log buffer and configures slightly smaller buffer of
messages and errors. This can be helpful
http://projects.theforeman.org/issues/14317

### DIFF
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -69,6 +69,8 @@ cat >/etc/foreman-proxy/settings.yml <<'CFG'
 :https_port: 8443
 :log_file: SYSLOG
 :log_level: DEBUG
+:log_buffer: 1000
+:log_buffer_errors: 500
 CFG
 
 cat >/etc/foreman-proxy/settings.d/discovery_image.yml <<'CFG'
@@ -83,6 +85,11 @@ cat >/etc/foreman-proxy/settings.d/bmc.yml <<'CFG'
 CFG
 
 cat >/etc/foreman-proxy/settings.d/facts.yml <<'CFG'
+---
+:enabled: true
+CFG
+
+cat >/etc/foreman-proxy/settings.d/logs.yml <<'CFG'
 ---
 :enabled: true
 CFG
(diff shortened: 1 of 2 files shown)


stbenjam:
Hm, should the factory really be used directly like this? The "standard" interface for logs has been to include/extend `::Proxy::Log` and then call the logger instance.  Is there any way we could do that here? The problem I'm also having is that I'm also not clear where this discovery.rb gets loaded in the context of the proxy.

lzap:
The method is public and supposed to be used, that's what I do. I don't see any benefit in extending/including it.

What you recommend then? I really don't want to just drop the messages. Syslog can only be opened once and proxy runs daemonized where STDOUT goes to `/dev/null`.

stbenjam:
I guess it's fine. What about my second question? How is this loaded in the context of the proxy?

lzap:
Via `Facter.facts.all` (pseudo code, don't remember the method name).

lzap:
The problem I see if I remove all the logging from utility methods which are used in custom facts is when we will be working on them later on, we can forget and add a log. And we won't catch it immediately, only after image is rebuild...

stbenjam:
My question still isn't answered, I'd rather not merge something that I don't understand how it works.

Is it not clear?

How does the code in /usr/lib64/ruby/vendor_ruby/discovery.rb end up the Smart Proxy?  Is it being required somewhere? I looked in the code base and I just can't find it.

lzap:
Oh sorry:

Smart Proxy -> Facter -> /usr/share/fdi/facts/* -> requires -> vendor_ruby/discovery.rb

stbenjam:
Oh I see, thanks!

stbenjam:
Merged as c4631779758166409e567b334aca8ea23ffbc098, thanks @lzap!

PR #71, opened 2016-04-06
Fixes #14494 - made texts branding-friendly

lzap:
@stbenjam I am removing all Foreman terms and using just Server instead
"Foreman Server" which is downstream-friendly. Filed a ticket to add full i18n
support but that's not on the table now. I don't even know if newt supports
UNICODE.

### DIFF
--- a/root/usr/lib64/ruby/vendor_ruby/discovery.rb
+++ b/root/usr/lib64/ruby/vendor_ruby/discovery.rb
@@ -158,7 +158,7 @@ end
 
 def upload(uri = discover_server, type = proxy_type, custom_facts = {})
   unless uri
-    log_err "Could not determine Foreman instance, add foreman.url or proxy.url kernel command parameter"
+    log_err "Could not determine instance type, add foreman.url or proxy.url kernel command parameter"
     return
   end
   unless uri.is_a? URI
@@ -166,10 +166,10 @@ def upload(uri = discover_server, type = proxy_type, custom_facts = {})
     return
   end
   if uri.host.nil? or uri.port.nil?
-    log_err "Foreman URI host or port was not specified, cannot continue"
+    log_err "Server or proxy URI host or port was not specified, cannot continue"
     return
   end
-  log_msg "Registering host with Foreman (#{uri})"
+  log_msg "Registering host at (#{uri})"
   http = Net::HTTP.new(uri.host, uri.port)
   if uri.scheme == 'https' then
     http.use_ssl = true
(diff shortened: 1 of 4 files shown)


ShimShtein:
Can we add support for [gettext](https://github.com/grosser/fast_gettext) here? It will be standardized and maybe we even make it translatable.

lzap:
Sure, the ticket already have associated one for proper i18n solution, this patch only makes it more neutral. I will followup with i18n solution soon, I would like not to block 6.2 with it, there will be build (extractions) and packaging changes necessary, maybe new dependency package on the image, thus hidden dragons I would like to avoid. Completely agree with you...

For the record the ticket is http://projects.theforeman.org/issues/14495

lzap:
If there are no objections to merge this, I will followup with i18n next week.

root/usr/lib64/ruby/vendor_ruby/discovery.rb
```diff
@@ -158,7 +158,7 @@ def write_tui result = 'success', code, body
 
 def upload(uri = discover_server, type = proxy_type, custom_facts = {})
   unless uri
-    log_err "Could not determine Foreman instance, add foreman.url or proxy.url kernel command parameter"
+    log_err "Could not determine instance type, add foreman.url or proxy.url kernel command parameter"
```
orrabin: As @ohadlevy  mentioned on irc we still have foreman.url and proxy.url here and we need them to determine the instance type.
Should we change it? just the `foreman.url`? keep it until the i18n change?

root/usr/lib64/ruby/vendor_ruby/discovery.rb
```diff
@@ -166,10 +166,10 @@ def upload(uri = discover_server, type = proxy_type, custom_facts = {})
     return
   end
   if uri.host.nil? or uri.port.nil?
-    log_err "Foreman URI host or port was not specified, cannot continue"
+    log_err "Server or proxy URI host or port was not specified, cannot continue"
```
orrabin: I think `proxy` is also problematic but not sure you can use anything else here.

lzap:
I have just pushed additional commit to this PR, so you can test both. The initial one is only meant for easy downstream merging, the latter is full features i18n support - to switch locale provide "locale=cs_CZ" command line option.

root/usr/lib64/ruby/vendor_ruby/discovery.rb
lzap: This and above are log output, users unlikely are facing this often, therefore I don't think it matters much. Those strings only appear in system log and only on errors.

orrabin: Then why change them at all?
I'm fine with leaving it either way and changing it only if we also change the url to something branded or a neutral name that works for both upstream and downstream.

lzap: Do you propose to change the name of the `foreman.url` option itself?

orrabin: Ignore my last comment, it's too confusing.

root/usr/lib64/ruby/vendor_ruby/discovery.rb
lzap: I don't follow. Branding is one thing, but I'd rather stick with configuration options. We are also not renaming all our "foreman_xyz" settings in core. I'd keep it.

lzap:
@orrabin the missing dependency is now in nightly

lzap:
Removed the i18n part.

stbenjam:
Thanks! Merged as eac3d6d424eb9bdebc1d35116c47623c7e649eb7

lzap:
Thanks, the i18n part filed as https://github.com/theforeman/foreman-discovery-image/pull/72

PR #73, opened 2016-05-23
Fixes #15138 - start proxy when IPv4 is not present

lzap:
When DHCP fails to deliver an IPv4 address (or on networks without DHCP),
foreman-proxy service might fail to start because HTTPS X509 CN might be empty
(openssl fails to generate cert without CN).

This patch adds a static string "discovered" there so users can still use
discovered node via HTTP. To test this, disable DHCP and start the image,
foreman-proxy should be up and running responding both on 8448 (HTTP) and 8443
(HTTPS with this common name - certificate should fail to validate of course).


dLobatog:
Merged as 0e9e0ba002fb12a6d305c3a5ec26d2e432de8619, thanks @lzap

PR #76, opened 2016-06-24
Fixes #9945 - add VLAN support for primary interface

lzap:
This will work in both PXE and PXE-less modes. Use `fdi.vlan.primary=10` to set
VLAN ID to 10 for the primary interface via kernel command line or use new VLAN
text option on the Primary Interface screen to configure it in PXE-less.

### DIFF
--- a/root/usr/bin/nm-configure
+++ b/root/usr/bin/nm-configure
@@ -6,6 +6,7 @@ bootif="$(normalizeHwAddr "${BOOTMAC}")"
 mac=${2:-$bootif}
 uuid=$(/usr/bin/uuid)
 timeout=${KCL_FDI_DHCP_TIMEOUT:-300}
+ctype="802-3-ethernet"
 
 correct_owner_perms() {
   chown root:root $1
@@ -14,8 +15,8 @@ correct_owner_perms() {
 
 usage() {
   echo "Usage: "
-  echo " $0 primary [MAC_ADDRESS]"
-  echo " $0 primary-static MAC_ADDRESS IP4_ADDR/CIDR IP4_GW IP4_DNS"
+  echo " $0 primary MAC_ADDRESS [VPN_ID]"
+  echo " $0 primary-static MAC_ADDRESS IP4_ADDR/CIDR IP4_GW IP4_DNS [VPN_ID]"
   echo " $0 secondary MAC_ADDRESS AUTOCONNECT (true/false)"
 }
 
@@ -25,12 +26,14 @@ if [[ "$1" == "primary-static" ]]; then
   ip=$3
   gw=$4
   dns=$5
+  vlanid=${6:-$KCL_FDI_VLAN_PRIMARY}
+  [[ "$vlanid" != "" ]] && ctype="vlan"
   script=/etc/NetworkManager/system-connections/primary
   cat > $script <<EONS
 [connection]
 id=primary
 uuid=$uuid
-type=802-3-ethernet
+type=$ctype
 autoconnect=true
 autoconnect-priority=1
 [802-3-ethernet]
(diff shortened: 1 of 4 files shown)


lzap:
Scratchbuild available at: http://downloads.theforeman.org/discovery/lzap/scratch/ for testing.

stbenjam:
@lzap Please rebase

lzap:
Done! Thanks for reviews.

root/usr/bin/nm-configure
```diff
@@ -42,15 +45,19 @@ gateway=$gw
 dns=$dns;
 [ipv6]
 method=ignore
+[vlan]
+id=$vlanid
 EONS
   correct_owner_perms $script
 elif [[ "$1" == "primary" ]]; then
+  vlanid=${3:-$KCL_FDI_VLAN_PRIMARY}
```
stbenjam: Shouldn't this be $6?

stbenjam: Oh disregard, I see - it's $3 in the primary care

stbenjam:
@lzap Everything looks good at first glance.  I wanted to give it a try on PXEless, but after building the image out of this branch, I'm getting some FastGetText errors.  That seems to be hiding the actual exception message - but it's blowing up in the main_loop: https://github.com/theforeman/foreman-discovery-image/blob/master/root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb#L212

vlan

Any ideas?

lzap:
I18n should have been initialized in https://github.com/theforeman/foreman-discovery-image/blob/master/root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb#L124-L137

Do you see something in system log? E.g. error during initialization which is logged (and ignored as I thought it won't hurt). Can you see `/usr/share/locale/fdi` content?

lzap:
Rebased, turned out the i18n issues were regression. This was fixed now.

dLobatog:
Thanks @lzap & @stbenjam

kalistuf:
Hello
@lzap , I have an issue with the vlan option  : when the server is deployed in PXE-Less mode and static network with a primary interface tagged with a vlan, the foreman plugin (server) discover the host and put the primary interface tag/fact to the physical interface (not the vlan interface) and return the error : Unable to assign subnet, primary interface is missing IP address. 
if I modify this line in the foreman discovery plugin :
https://github.com/theforeman/foreman_discovery/blob/a6001b102eb779b02dac2544f47818165cea85af/app/services/foreman_discovery/fact_parser.rb#L6
by  : 
detected = interfaces.detect { |_, values| values[:macaddress].try(:downcase) == bootif_mac && values[:ipaddress].present?}
foreman discovery plugin detect the right interface (<interface>.<vlanid> but there is no ip address found and it is not possible to deploy automatically the server. 
the server appear in the "host discovered" page but it not possible to deploy manually the server because of the Location is grey.
what is missing ?

lzap:
Hello, can you paste:

    FACTERLIB=/usr/share/fdi/facts/ facter --json

to see all facts reported after you turn on network and attempt to discover?

What subnets do you have defined and do they have locations assigned? Org/Loc is determined from Subnet the discovery host belongs to, if not found than default (first) org/loc is used.

kalistuf:
[facts.txt](https://github.com/theforeman/foreman-discovery-image/files/1243551/facts.txt)
the facts are anonymized.
I have an network configured with the vlan id and an org adns location  is assigned to the network.

lzap:
Please file a RedMine issue, I tried your patch and I am still unable to discover the server. In Foreman Core NIC importer code it looks like there is a bug or missing feature of importing VLAN primary interface - IP is left blank. This needs to be fixed first in core I believe. So perhaps two issues - one for core, one for discovery.

PR #77, opened 2016-06-24
Fixes #15517 - root password is not sent to journal

lzap:
Asking for quick review and merge @stbenjam or @dLobato or anybody.


stbenjam:
Thanks @lzap

PR #79, opened 2016-09-02
fixes #16420 - report the foreman server/proxy used to register

sean797:


lzap:
Thanks that works, note there it a bug in `master` - gettext fails to initialize. I have a PR that fixes it: https://github.com/theforeman/foreman-discovery-image/pull/79 I will merge this patch and rebase the other on top of and make a scratch build of the image so you can test it.

lzap:
Merged as a928c45645837965bf9001d9fc8cdd0e00ca9f43, thank you!

PR #80, opened 2016-09-06
Fixes #16450 - fixed foreman port check

lzap:
Port was never nil, but 80 when not provided so the if statement was not
working. We should not enforced, so I also made it just a warning message box.


dLobatog:
Merged as df09a83a354732b4f2ea48e8f009764c002b3945, thanks @lzap!

dLobatog:
It's almost funny this PR number is 80

PR #81, opened 2016-09-13
Fixes #16526 - added biosdevname package

lzap:
Unless you test this on DELL system, you can't really see any difference.
Ensure the presence of biosdevname command and that's it.


dLobatog:
Thanks @lzap !

PR #84, opened 2017-03-13
Fixes #18148 - send DHCP and DNS-PTR facts

lzap:
Adds list of extra facts - all extracted from NetworkManager, connection
`primary` which is always present (in PXE less mode after network is
initialized). Example output can be seen here:
http://projects.theforeman.org/issues/18148#note-6

Adds also extra fact called `nmprimary_ptr` which is PTR record of primary IPv4
IP address.

The idea (and requests from users) is to use external DHCP domain and hostnames
(`nmprimary_dhcp4_option_domain_name` and `nmprimary_dhcp4_option_host_name`)
in provisioning templates or as hostname (`discovery_fact` setting). The same
goes for `nmprimary_ptr` for those who prefer to use DNS PTR external records.

### DIFF
--- a/root/usr/share/fdi/facts/discovery-facts.rb
+++ b/root/usr/share/fdi/facts/discovery-facts.rb
@@ -18,6 +18,7 @@
 # MA  02110-1301, USA.  A copy of the GNU General Public License is
 # also available at http://www.gnu.org/copyleft/gpl.html.
 
+require 'resolv'
 require 'facter/util/ip'
 require '/usr/lib64/ruby/vendor_ruby/discovery.rb'
 
@@ -158,6 +159,49 @@ if has_ipmi
     end
     attributes.each do |fact, value|
       add_ipmi_facts fact, value, n
+    end
+  end
+end
+
+# NetworkManager details (e.g. nmprimary_dhcp4_option_domain_name)
+nmout = Facter::Util::Resolution.exec("nmcli -t con show primary 2>/dev/null")
+nmout.each_line do |x|
+  elements = x.split(":", 2)
+  name = elements.first.downcase.sub(/\[\d+\]$/,"")
+  if name =~ /dhcp.\.option/
+    dhcp_elems = elements.last.split(/\s*=\s*/, 2)
+    name += '_' + dhcp_elems.first
+    value = dhcp_elems.last
+  else
+    value = elements.last
+  end
+  name = "nmprimary_" + name.tr('.', '_')
+  Facter.add(name) do
+    setcode do
+      value.chomp
+    end
+  end
+end
+
+# Create DHCP FQDN helper facts
+Facter.add("nmprimary_dhcp4_option_fqdn") do
+  setcode do
+    if Facter.value("nmprimary_dhcp4_option_host_name") && Facter.value("nmprimary_dhcp4_option_domain_name")
+      Facter.value("nmprimary_dhcp4_option_host_name") + '.' + Facter.value("nmprimary_dhcp4_option_domain_name")
+    else
+      nil
+    end
+  end
+end
+
+# Primary interface IP PTR DNS record
+Facter.add("nmprimary_ptr") do
+  setcode do
+    ip = Facter.value("nmprimary_ip4_address").scan(/\d+\.\d+\.\d+\.\d+/).first.strip rescue nil
+    if ip.nil?
+      nil
+    else
+      Resolv.new.getname(ip) rescue nil
     end
   end
 end


lzap:
Pushed one change - since concatenation in `discovery_facts` global setting is not possible and this is very useful, I created extra helper fact `nmprimary_dhcp4_option_fqdn`. @timogoebel opinions?

lzap:
When testing please create PTR records for all IPs from DHCP discovery pool and check `nmprimary_ptr` and make reservation for MAC address and check `nmprimary_dhcp4_option_fqdn`. Also test PXE less mode. Thanks!

root/usr/share/fdi/facts/discovery-facts.rb
```diff
@@ -161,3 +162,41 @@ def add_ipmi_facts fact, value, channel = nil
...
+# Create DHCP FQDN helper facts
+Facter.add("nmprimary_dhcp4_option_fqdn") do
+  setcode do
+    nmfacts["nmprimary_dhcp4_option_host_name"] + '.' + nmfacts["nmprimary_dhcp4_option_domain_name"]
+  end
+end
+
+# Primary interface IP PTR DNS record (or unique random name if not found)
+ip = nmfacts["nmprimary_ip4_address"].scan(/\d+\.\d+\.\d+\.\d+/).first.strip rescue '0.0.0.0'
+value = Resolv.new.getname(ip) rescue "noptr#{rand(99999999)}"
```
timogoebel: Why fallback to a random name? I think, it would be better not to set the fact at all if the host doesn't have a ptr record.

lzap: Fixed, my thinking was if user sets this as primary fact for hostname that it could cause trouble, but this is cleaner I get it.

root/usr/share/fdi/facts/discovery-facts.rb
```diff
@@ -161,3 +162,41 @@ def add_ipmi_facts fact, value, channel = nil
...
+    setcode do
+      value
+    end
+  end
+end
+
+# Create DHCP FQDN helper facts
+Facter.add("nmprimary_dhcp4_option_fqdn") do
+  setcode do
+    nmfacts["nmprimary_dhcp4_option_host_name"] + '.' + nmfacts["nmprimary_dhcp4_option_domain_name"]
```
timogoebel: This gives me this:
```
Could not retrieve fact='nmprimary_dhcp4_option_fqdn', resolution='<anonymous>': undefined method `+' for nil:NilClass
```

root/usr/share/fdi/facts/discovery-facts.rb
```diff
@@ -161,3 +162,41 @@ def add_ipmi_facts fact, value, channel = nil
...
+  end
+end
+
+# Primary interface IP PTR DNS record
+ip = nmfacts["nmprimary_ip4_address"].scan(/\d+\.\d+\.\d+\.\d+/).first.strip rescue '0.0.0.0'
```
timogoebel: I think, it would be a lot more clearer to rescue `nil` here and check and only run the code below if the ip is not nil. Setting it to `0.0.0.0` is functional the same but might actually issue a dns query that has to come back with no result.

root/usr/share/fdi/facts/discovery-facts.rb
```diff
@@ -161,3 +162,41 @@ def add_ipmi_facts fact, value, channel = nil
...
+value = Resolv.new.getname(ip) rescue nil
+Facter.add("nmprimary_ptr") do
+  setcode do
+    value
```
timogoebel: If you move the resolving code here, you can save some lines of code.

```ruby
Facter.add("nmprimary_ptr") do
  setcode do
    Resolv.new.getname(ip) rescue nil
  end
end
```

Facts with `nil` as value are treated as empty.

lzap:
Thanks pushed changes.

root/usr/share/fdi/facts/discovery-facts.rb
```diff
@@ -161,3 +162,48 @@ def add_ipmi_facts fact, value, channel = nil
...
+    name += '_' + dhcp_elems.first
+    value = dhcp_elems.last
+  else
+    value = elements.last
+  end
+  name = "nmprimary_" + name.tr('.', '_')
+  nmfacts[name] = value
+  Facter.add(name) do
+    setcode do
+      value
```
timogoebel: I'd suggest using `.stip` on these values or I get

```
facter nmprimary_dhcp4_option_fqdn
abc
.example.com```

**Comment by lzap** [comment]
All done, these were good points. I don't know facter at all to be honest. Surprise! ;-)

File: root/usr/share/fdi/facts/discovery-facts.rb [hunk 6]
```diff
@@ -161,3 +162,46 @@ def add_ipmi_facts fact, value, channel = nil
...
+    setcode do
+      value.chomp
+    end
+  end
+end
+
+# Create DHCP FQDN helper facts
+Facter.add("nmprimary_dhcp4_option_fqdn") do
+  setcode do
+    if Facter.value["nmprimary_dhcp4_option_host_name"] && Facter.value["nmprimary_dhcp4_option_domain_name"]
```
timogoebel: You need to use round brackets here. `Facter.value()`. `Facter.value[]` raises an error for me.
Same below.

lzap: Oh my bad, fixed sorry.

timogoebel:
thanks @lzap !
Looking forward to the plugin implementation of this.

lzap:
Thanks a lot @timogoebel *big help* here, my PRs tool so much long to get merged. Well I don't plan any code on the plugin side, you can use these facts in discovery_fact setting and templates already. Anything else you had on your mind?

timogoebel:
@lzap :
> my PRs tool so much long to get merged

Yeah. I know the feeling. Everybody has too much to do I guess.

> Anything else you had on your mind?

Actually, no. I just had this confused.
This reminds me though, that I'd also like to have the PTR of the ipmi card. E.g. at work the PTR is set by the datacenter team when they rack a server. So you could determine what server just checked into discovery. I'll make a ticket as a reminder.

PR #85, opened 2017-04-07
Fixes #19220 - properly handle SIGTERM in menu

lzap:
Looks like menu is blocking it, does not respond to TERM and systemd waits 1:30
minute to send TERM. The signal had a handler but it just cleared screen and it
never exited the process. Whoopsy.


lzap:
@timogoebel would love to have this one in 1.15, I will respin the image... so annoying one.

timogoebel:
@lzap : Building the image to test this now.

PR #87, opened 2017-05-22
Dns srv record fix

barrymw:
discovery.rb update as outlined on the Red Hat community discussion: [Discovery Plugin / Capsule and DNS SRV Record](https://access.redhat.com/discussions/2490321#comment-1170261)


lzap:
Hello, thank you for your contribution!
I actually replaced your proposal with mine in this case, because I believe this should be configurable:
https://github.com/theforeman/foreman-discovery-image/pull/88
I will build you a test build ISO so you can test it in your environment quickly. See the PR for download link.

PR #91, opened 2017-06-12
Fixes #19984 - added fdi.countdown to unattended pxeless

lzap:
Slow networks workaround.


dLobatog:
Thanks, out of curiosity when did you need this?

timogoebel:
@lzap: Is this related to https://github.com/theforeman/foreman-discovery-image/pull/90?

lzap:
Sorta, this is the same thing but when you are using PXE less unattended mode (you put fdi.auto=1 onto cmdline).

lzap:
So this was suggested by our customer, you can only run into this on slow DHCP networks or when NM is slow with net init (drivers, hardware).

PR #92, opened 2017-06-21
Adds tcpdump to fdi image

sean797:
Small thing, I couldn't think of any others that would be useful.. Maybe someone can


sean797:
Thanks @dLobatog just for debugging network issues with FDI. I had a problem with FDI receiving the kexec 'command', tcpdump would have been useful debugging it.

dLobatog:
Thanks again @sean797

PR #94, opened 2017-06-28
Fixes #20127,#20128 - nm autoloading and atomicity

lzap:
For easier testing, two patches in one PR, P1:

NetworkManager monitors configuration profile changes and automatically
reloads configurations. This patch makes sure it's atomic file system
change (move) rather than create/modify owner/perms which confused NM
issuing warnings about insecure permissions.

P2 is all about autoloading of nm, we have configured NM to detect profile changes and reload automatically (see 22-discovery.ks) but then we issue "down" and "reload" commands via nmcli. We need to pick either one or the other approach - it looks like keeping autoloading (configuration file monitoring) is better for PXE startup, therefore I removed the "nmcli con down" command (which also fixes 19950 issue) and the code now relies heavily on NM ability to detect the change.

Also we touched the configuration file three times (create, chown, chmod) which caused warnings and extra reloading in NM, so I changed the code to atomically move the file to the destination.

During testing I hit an issue when HTTPS SSL certificate for proxy (its executed during startup of the service) was started *before* IP address is ackquired (nm-online does not wait for IP4 DHCP to complete), therefore I added a code that waits one minute and every second it attempts to read IPv4 address from NM - when it appears, the certificate is generated with correct COMMON NAME (of the given IP), otherwise it falls back to "discovered" string.

### DIFF
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -15,8 +15,8 @@ cat > /etc/NetworkManager/NetworkManager.conf <<'NM'
 [main]
 monitor-connection-files=yes
 no-auto-default=*
-#[logging]
-#level=DEBUG
+[logging]
+level=DEBUG
 NM
 cat > /etc/udev/rules.d/81-nm-prepare.rules <<'UDEV'
 ACTION=="add", SUBSYSTEM=="net", NAME!="lo", RUN+="/usr/bin/systemd-cat -t nm-prepare /usr/bin/nm-prepare %k"
--- a/root/usr/bin/generate-proxy-cert
+++ b/root/usr/bin/generate-proxy-cert
@@ -10,8 +10,14 @@ exportKCL
 
 DIR=/etc/foreman-proxy
 DAYS=${KCL_FDI_PROXY_CERT_DAYS:-999}
-
-COMMON_NAME=$(nmcli -t -f IP4.ADDRESS con show primary 2>/dev/null | cut -f2 -d: | cut -f1 -d/)
+WAIT=${KCL_FDI_NMWAIT:-120}
+SECONDS=0
+while (( SECONDS < 60 )); do
+  sleep 1
+  COMMON_NAME=$(nmcli -w $WAIT -t -f IP4.ADDRESS con show primary 2>/dev/null | cut -f2 -d: | cut -f1 -d/)
+  [ ! -z "$COMMON_NAME" ] && break
+  logger "Waiting for IP4 address to generate SSL cert ($SECONDS)"
+done
 
 # Don't fail when IP address was not provided (HTTP can be still used).
 [ -z "$COMMON_NAME" ] && COMMON_NAME=discovered
--- a/root/usr/bin/nm-configure
+++ b/root/usr/bin/nm-configure
@@ -8,9 +8,10 @@ uuid=$(/usr/bin/uuid)
 timeout=${KCL_FDI_DHCP_TIMEOUT:-300}
 ctype="802-3-ethernet"
 
-correct_owner_perms() {
+deploy_config() {
   chown root:root $1
   chmod 600 $1
+  mv $1 $2
 }
 
 usage() {
@@ -22,6 +23,12 @@ usage() {
 
 [[ -z "$mac" ]] && echo "MAC address was not provided or detected, leaving unconfigured" && exit 0
 
+function cleanup() {
+  [ -f $TMP_CFG ] && rm -f $TMP_CFG
+}
+TMP_CFG=$(mktemp)
+trap cleanup EXIT SIGINT SIGTERM
+
 if [[ "$1" == "primary-static" ]]; then
   ip=$3
   gw=$4
@@ -29,13 +36,14 @@ if [[ "$1" == "primary-static" ]]; then
   vlanid=${6:-$KCL_FDI_VLAN_PRIMARY}
   [[ "$vlanid" != "" ]] && ctype="vlan"
   script=/etc/NetworkManager/system-connections/primary
-  cat > $script <<EONS
+  cat > $TMP_CFG <<EONS
 [connection]
 id=primary
 uuid=$uuid
 type=$ctype
 autoconnect=true
 autoconnect-priority=1
+may-fail=false
 [802-3-ethernet]
 mac-address=$mac
 [ipv4]
@@ -48,18 +56,19 @@ method=ignore
 [vlan]
 id=$vlanid
 EONS
-  correct_owner_perms $script
+  deploy_config $TMP_CFG $script
 elif [[ "$1" == "primary" ]]; then
   vlanid=${3:-$KCL_FDI_VLAN_PRIMARY}
   [[ "$vlanid" != "" ]] && ctype="vlan"
   script=/etc/NetworkManager/system-connections/primary
-  cat > $script <<EONS
+  cat > $TMP_CFG <<EONS
 [connection]
 id=primary
 uuid=$uuid
 type=$ctype
 autoconnect=true
 autoconnect-priority=1
+may-fail=false
 [802-3-ethernet]
 mac-address=$mac
 [ipv4]
@@ -71,17 +80,18 @@ method=ignore
 [vlan]
 id=$vlanid
 EONS
-  correct_owner_perms $script
+  deploy_config $TMP_CFG $script
 elif [[ "$1" == "secondary" ]]; then
   autoconnect=$3
   id=secondary-$mac
   script=/etc/NetworkManager/system-connections/$id
-  cat > $script <<EONS
+  cat > $TMP_CFG <<EONS
 [connection]
 id=$id
 uuid=$uuid
 type=802-3-ethernet
 autoconnect=$autoconnect
+may-fail=true
 [802-3-ethernet]
 mac-address=$mac
 [ipv4]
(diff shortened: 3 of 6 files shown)


lzap:
Rebased @dLobatog 

One more thing, this patch addes `fdi.nmwait` which passes `--wait` option, it's by default 90 for up and 10 for down commands, initially I though't the 10 seconds is the issue for RM 20128 but I am keeping this setting - might be useful when NM starts slowly.

lzap:
To test perform:

* PXE-less DHCP discovery
* PXE-less w/o DHCP
* fully unattended PXE-less discovery

Verify smart-proxy is running (refresh facts) in all cases.

20-packages.ks
```diff
@@ -94,17 +94,10 @@ system-storage-manager
...
-# selinux toolchain of policycoreutils, libsemanage, ustr
+# selinux toolchain of policycoreutils, libsemanage (libselinux is needed tho)
 -policycoreutils
 -checkpolicy
 -selinux-policy*
--libselinux-python
--libselinux
 
 # Things it would be nice to loose
 -fedora-logos
```
lzap: Yum started to complain recently without this change, likely an update in CentOS.

22-discovery.ks
```diff
@@ -15,8 +15,8 @@ cat > /etc/NetworkManager/NetworkManager.conf <<'NM'
 [main]
 monitor-connection-files=yes
 no-auto-default=*
-#[logging]
-#level=DEBUG
+[logging]
+level=DEBUG
```
lzap: Won't hurt, the level of detail is decent yet not spamming with trace messages. Journald is configured to use 20MB of memory for volatile backlog.

build-livecd-root
```diff
@@ -81,7 +81,7 @@ EOF
 sha256sum initrd0.img vmlinuz0 > SHA256SUM
 
 cd ..
-tarball=$srcdir/fdi-image-${last_tag#release-}.tar
+tarball=$destdir/fdi-image-${last_tag#release-}.tar
 tar cf $tarball fdi-image/
```
lzap: I added second parameter to this script which allows you to set OUTPUT to different directory.

root/usr/bin/generate-proxy-cert
```diff
@@ -10,8 +10,14 @@ exportKCL
...
+WAIT=${KCL_FDI_NMWAIT:-120}
+SECONDS=0
+while (( SECONDS < 60 )); do
+  sleep 1
+  COMMON_NAME=$(nmcli -w $WAIT -t -f IP4.ADDRESS con show primary 2>/dev/null | cut -f2 -d: | cut -f1 -d/)
+  [ ! -z "$COMMON_NAME" ] && break
+  logger "Waiting for IP4 address to generate SSL cert ($SECONDS)"
+done
 
 # Don't fail when IP address was not provided (HTTP can be still used).
```
lzap: Before proxy can start, it needs SSL cert and it needs IP address for Common Name in order it to be valid. That's why it needs to wait. There is not reasonable way to detect when IP address was assigned other than this.

root/usr/bin/nm-configure
```diff
@@ -22,20 +23,27 @@ usage() {
 
 [[ -z "$mac" ]] && echo "MAC address was not provided or detected, leaving unconfigured" && exit 0
 
+function cleanup() {
+  [ -f $TMP_CFG ] && rm -f $TMP_CFG
+}
+TMP_CFG=$(mktemp)
+trap cleanup EXIT SIGINT SIGTERM
+
```
lzap: Atomicity - write to temp, chmod/chown and then move it.

root/usr/bin/nm-configure
```diff
@@ -48,18 +56,19 @@ method=ignore
...
   script=/etc/NetworkManager/system-connections/primary
-  cat > $script <<EONS
+  cat > $TMP_CFG <<EONS
 [connection]
 id=primary
 uuid=$uuid
 type=$ctype
 autoconnect=true
 autoconnect-priority=1
+may-fail=false
```
lzap: Not much relevant, I was just browsing docs trying to find something and found this option.

root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
```diff
@@ -76,16 +76,17 @@ def start_discovery_service
...
 
 def configure_network static, mac, ip=nil, gw=nil, dns=nil, vlan=nil
-  command("systemctl stop foreman-proxy")
+  command("systemctl stop foreman-proxy", false)
   if static
     command("nm-configure primary-static '#{mac}' '#{ip}' '#{gw}' '#{dns}' '#{vlan}'")
   else
     command("nm-configure primary '#{mac}' '#{vlan}'")
   end
-  result = command("nmcli connection up primary", false)
```
lzap: Note the "up" command is also not necessary at all, everything is automatic.

Unfortunately we are not reliably tell here if configuration was success or not, NM does the job and `nm-online` usually returns 0 very quickly (when NM daemon is up and at least one "must-not-fail" interface is starting = success).

lzap:
@dLobatog can you please test this scenario as well (from https://bugzilla.redhat.com/show_bug.cgi?id=1465643):

1. create a guest with 2 or more cpus
2. boot the foreman-discovery-image
3. select the nic
4. choose manual
5. set network parameters
6. the foreman-discovery-image gets broken, since it was unable to bringup the network

lzap:
We need to test this with fdi.pxauto=1 as well.

lzap:
@dLobatog any chance of merging this? I will test this tomorrow - all the workflows so you can only focus on code review. I need to be sure that PXE-less workflow works before I do 3.4.1 release (downstream one).

rplevka:
I've tested this on following scnearios:

| Mode  | DHCP | provision with kexec | result | comment |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| PXE  | true  | false | PASS |
| PXE less  | true | true | PASS | attended mode (used dhcp discovery)
| PXE less  | false | true | PASS | attended mode (nw details specified manually using TUI)
| PXE less  | false | true  | PASS | unattended mode, with remastered image: `discovery-remaster fdi-bootable-3.4.0.iso 'proxy.url=https://10.16.70.13:443 proxy.type=foreman fdi.pxip=192.168.100.123/24 fdi.pxgw=192.168.100.1 fdi.pxdns=192.168.100.1 fdi.pxfactname1=fdi fdi.pxfactvalue1=3.4.1-remastered fdi.pxmac=52:54:00:83:79:47 fdi.pxauto=1'`
| PXE less  | true | true | PASS | unattended mode, with remastered image: `discovery-remaster fdi-bootable-3.4.0.iso 'proxy.url=https://10.16.70.13:443 proxy.type=foreman fdi.pxip=192.168.100.123/24 fdi.pxgw=192.168.100.1 fdi.pxdns=192.168.100.1 fdi.pxmac=52:54:00:83:79:47 fdi.pxauto=1'`
I've used VM and assigned it ~20 cores.

root/usr/bin/nm-configure
dLobatog: As far as I can tell from here, this should go in the `[ipv4]` section, shouldn't it? https://lazka.github.io/pgi-docs/NetworkManager-1.0/classes/SettingIP4Config.html

root/usr/bin/nm-configure
```diff
@@ -71,17 +80,18 @@ method=ignore
...
   id=secondary-$mac
   script=/etc/NetworkManager/system-connections/$id
-  cat > $script <<EONS
+  cat > $TMP_CFG <<EONS
 [connection]
 id=$id
 uuid=$uuid
 type=802-3-ethernet
 autoconnect=$autoconnect
+may-fail=true
```
dLobatog: Ditto, `[ipv4]`

root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
```diff
@@ -76,16 +76,17 @@ def start_discovery_service
...
-  command("nm-online -s -q --timeout=45") unless static
+  # NetworkManager monitors config files and auto connects immediately
+  wait = cmdline('fdi.nmwait', 120)
+  result = command("nm-online -s -q --timeout=#{wait}")
   # restarting proxy with regenerated SSL self-signed cert
-  command("systemctl start foreman-proxy") if result
```
dLobatog: Why removing this if?

root/usr/bin/nm-configure
lzap: Ok these are irrelevant to the PR, I am removing - I do not want to cause any regressions now. It was dumb idea to introduce it here, sorry.

root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
lzap: It always returns true. More than that, `nm-online` is actually *useless* tool, it does return even when network is not yet up. I had to workaround this above - via the wait loop.

After *hours and hours* of googling and trying, I am pretty confident to say - there is no way in systemd/NetworkManager to wait until network is initialized via DHCP. At least `nm-online` wont you get that. I think that's on purpose, they do not want services to be waiting for that. We need this tho, because we need the IP for SSL cert.

lzap:
Was this all? Rebased, many thanks for looking into it.

dLobatog:
Merged, thanks @lzap

PR #95, opened 2017-06-29
Fixes #20163 - added libselinux and missing deps

lzap:
Taken out from

https://github.com/theforeman/foreman-discovery-image/pull/94

we need to add some more dependencies which is unfortunate, ISO file will grow a bit again.

In the future we want to try Atomic as our base.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ root/opt/extension/facts/*
 root/opt/extension/lib/ruby/*
 root/opt/extension/lib/*
 root/opt/extension/bin/*
+fdi-rhel7.ks
+00-repos-rhel7.ks
--- a/20-packages.ks
+++ b/20-packages.ks
@@ -77,9 +77,14 @@ e2fsprogs
 bzip2
 system-storage-manager
 
-#
-# Packages to Remove
-#
+######################
+# Packages to Remove #
+######################
+
+# Red Hat Enteprise Linux subscription tool
+-subscription-manager
+
+# Generic and wireless tools
 -prelink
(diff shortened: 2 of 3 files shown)


lzap:
I will add the updated script as well.

lzap:
Done @dLobatog or @timogoebel mind taking a look? I need this downstream, does not build against RHEL or CentOS 7.4 beta.

.gitignore
```diff
@@ -11,3 +11,5 @@ root/opt/extension/facts/*
 root/opt/extension/lib/ruby/*
 root/opt/extension/lib/*
 root/opt/extension/bin/*
+fdi-rhel7.ks
+00-repos-rhel7.ks
```
lzap: These files are not commited as they contain internal HTTP links into brew and won't be useful.

20-packages.ks
```diff
@@ -77,9 +77,14 @@ e2fsprogs
...
 
-#
-# Packages to Remove
-#
+######################
+# Packages to Remove #
+######################
+
+# Red Hat Enteprise Linux subscription tool
+-subscription-manager
```
lzap: We used to have it downstream but now it needs more dependencies so let's get rid of it.

lzap:
Done @dLobatog or @timogoebel mind taking a look? I need this downstream, does not build against RHEL or CentOS 7.4 beta.

lzap:
FDI downstream build is currently blocked by this @ehelms

dLobatog:
Thanks @lzap !

PR #97, opened 2017-08-15
Fixes #20603 - Disable kernel messages on tty1

lzap:
After PXE-less boot, sign in onto the image and verify:

[lzap@box ~]$ sysctl kernel.printk
kernel.printk = 0       4       1       7

That the first column is zero so kernel will not print any messages on tty.


lzap:
This is a trivial patch I am merging as I am going to perform testing of develop.

PR #98, opened 2017-10-13
UX rename Proxy to Smary Proxy

sean797:
Small thing, but `Proxy` could be confused with HTTP proxy.


ares:
In fact in last upstream discussion I believe it was agreed that Foreman Proxy is the preferred name[1]. While we still have Smart Proxy all around, I think we should not add more 

[1] https://groups.google.com/forum/#!searchin/foreman-dev/smart$20or$20foreman$20proxy$20name%7Csort:relevance/foreman-dev/H-jbnDY5mhI/qHofd_dvBAAJ

timogoebel:
... how about just "Foreman Proxy"?

ares:
That's what I tried to suggest, sorry for not making myself clear :-)

sean797:
Thanks guys, I forgot about that thread :-) Updated!

lzap:
Thanks.

PR #99, opened 2018-01-02
Fixes #22112 - PXE-less DHCP first attempt fixed

lzap:
This patch aims fixing two issues. First, in PXE-less mode when DHCP configuration was selected, the first attempt always failed. The patch changes behavior of NetworkManager - it will not detect configuration file changes automatically anymore but it must be explicitly started via "nmcli c up" now (one exception is when NetworkManager starts up).

Second, when `fdi.dhcp_timeout` option was provided (defaults to 120) it did not have an effect. This was due to waiting loop, it was essentially fixed to 60 seconds. It turns out that -w / --wait options for nmcli has no effect in waiting for IP address. It is basically just a timeout for NetworkManager connection (DBUS) or other actions, now the loop respects the timeout value.

Please test both PXE less (with/without DHCP) and PXE workflows.

### DIFF
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -13,7 +13,7 @@ systemctl disable kdump.service
 echo " * configuring NetworkManager and udev/nm-prepare"
 cat > /etc/NetworkManager/NetworkManager.conf <<'NM'
 [main]
-monitor-connection-files=yes
+monitor-connection-files=no
 no-auto-default=*
 [logging]
 level=DEBUG
@@ -56,6 +56,7 @@ sed -i '/\[Unit\]/a ConditionPathExists=/etc/NetworkManager/system-connections/p
 sed -i '/\[Service\]/a EnvironmentFile=-/etc/default/discovery' /usr/lib/systemd/system/foreman-proxy.service
 sed -i '/\[Service\]/a ExecStartPre=/usr/bin/generate-proxy-cert' /usr/lib/systemd/system/foreman-proxy.service
 sed -i '/\[Service\]/a PermissionsStartOnly=true' /usr/lib/systemd/system/foreman-proxy.service
+sed -i '/\[Service\]/a TimeoutStartSec=9999' /usr/lib/systemd/system/foreman-proxy.service
 /sbin/usermod -a -G tty foreman-proxy
 
 cat >/etc/foreman-proxy/settings.yml <<'CFG'
--- a/root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
+++ b/root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
@@ -82,12 +82,12 @@ def configure_network static, mac, ip=nil, gw=nil, dns=nil, vlan=nil
   else
     command("nm-configure primary '#{mac}' '#{vlan}'")
   end
-  # NetworkManager monitors config files and auto connects immediately
   wait = cmdline('fdi.nmwait', 120)
-  result = command("nm-online -s -q --timeout=#{wait}")
-  # restarting proxy with regenerated SSL self-signed cert
-  command("systemctl start foreman-proxy")
-  result
+  command("nmcli -w #{wait} connection reload", false)
+  up_result = command("nmcli -w #{wait} connection up primary", false)
+  command("nm-online -s -q --timeout=#{wait}")
+  # wait for IPv4, generate SSL self-signed cert and start proxy
+  command("systemctl start foreman-proxy") && up_result
 end
 
 def perform_upload proxy_url, proxy_type, custom_facts
--- a/root/usr/bin/generate-proxy-cert
+++ b/root/usr/bin/generate-proxy-cert
@@ -10,13 +10,13 @@ exportKCL
 
 DIR=/etc/foreman-proxy
 DAYS=${KCL_FDI_PROXY_CERT_DAYS:-999}
-WAIT=${KCL_FDI_NMWAIT:-120}
+WAIT=${KCL_FDI_IPWAIT:-120}
 SECONDS=0
-while (( SECONDS < 60 )); do
+while (( SECONDS < $WAIT )); do
   sleep 1
-  COMMON_NAME=$(nmcli -w $WAIT -t -f IP4.ADDRESS con show primary 2>/dev/null | cut -f2 -d: | cut -f1 -d/)
+  COMMON_NAME=$(nmcli -w 1 -t -f IP4.ADDRESS con show primary 2>/dev/null | cut -f2 -d: | cut -f1 -d/)
   [ ! -z "$COMMON_NAME" ] && break
-  logger "Waiting for IP4 address to generate SSL cert ($SECONDS)"
+  logger "Waiting for IP4 address to generate SSL cert ($SECONDS/$WAIT)"
 done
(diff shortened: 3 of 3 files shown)


22-discovery.ks
```diff
@@ -22,6 +22,11 @@ cat > /etc/udev/rules.d/81-nm-prepare.rules <<'UDEV'
 ACTION=="add", SUBSYSTEM=="net", NAME!="lo", RUN+="/usr/bin/systemd-cat -t nm-prepare /usr/bin/nm-prepare %k"
 UDEV
 
+echo " * disable IPv6"
+cat > /etc/sysctl.d/40-ipv6.conf <<EOIP
+net.ipv6.conf.all.disable_ipv6 = 1
+EOIP
```
lzap: Also I noticed that IPv6 is disabled in nm-configure but not globally for PXE mode.

22-discovery.ks
```diff
@@ -22,6 +22,11 @@ cat > /etc/udev/rules.d/81-nm-prepare.rules <<'UDEV'
 ACTION=="add", SUBSYSTEM=="net", NAME!="lo", RUN+="/usr/bin/systemd-cat -t nm-prepare /usr/bin/nm-prepare %k"
 UDEV
 
+echo " * disable IPv6"
```
timogoebel: Why would you want to do this?

lzap: We already do this but per individual connections, I want this globally. Until we decide to support IPv6 in discovery.

lzap: You know what - removed from this patch, it's irrelevant here. But please note IPv6 is disabled on connection level atm.

22-discovery.ks
```diff
@@ -56,6 +56,7 @@ sed -i '/\[Unit\]/a ConditionPathExists=/etc/NetworkManager/system-connections/p
 sed -i '/\[Service\]/a EnvironmentFile=-/etc/default/discovery' /usr/lib/systemd/system/foreman-proxy.service
 sed -i '/\[Service\]/a ExecStartPre=/usr/bin/generate-proxy-cert' /usr/lib/systemd/system/foreman-proxy.service
 sed -i '/\[Service\]/a PermissionsStartOnly=true' /usr/lib/systemd/system/foreman-proxy.service
+sed -i '/\[Service\]/a TimeoutStartSec=9999' /usr/lib/systemd/system/foreman-proxy.service
```
lzap: This is to prevent killing foreman-proxy when it is waiting for IPv4. Although systemd supports "infinity" here, version in RHEL 7.4 does not have it, thus 9999.

root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
```diff
@@ -82,12 +82,12 @@ def configure_network static, mac, ip=nil, gw=nil, dns=nil, vlan=nil
...
   wait = cmdline('fdi.nmwait', 120)
-  result = command("nm-online -s -q --timeout=#{wait}")
-  # restarting proxy with regenerated SSL self-signed cert
+  command("nmcli -w #{wait} connection reload", false)
+  command("nmcli -w #{wait} connection up primary", false)
+  command("nm-online -s -q --timeout=#{wait}")
+  # wait for IPv4, generate SSL self-signed cert and start proxy
   command("systemctl start foreman-proxy")
-  result
 end
```
lzap: It is actually the very last command which is waiting for IP address to be available, all the others exit immediately.

lzap:
@timogoebel @ares @dLobatog guys review and testing appreciated this week, I need to hit Friday deadline with this. I will be likely filing two more but smaller patches.

timogoebel:
@lzap: I most likely won't be able to review this week, sorry.

root/usr/lib64/ruby/vendor_ruby/discovery/menu.rb
```diff
@@ -82,12 +82,12 @@ def configure_network static, mac, ip=nil, gw=nil, dns=nil, vlan=nil
...
-  # NetworkManager monitors config files and auto connects immediately
   wait = cmdline('fdi.nmwait', 120)
-  result = command("nm-online -s -q --timeout=#{wait}")
-  # restarting proxy with regenerated SSL self-signed cert
+  command("nmcli -w #{wait} connection reload", false)
+  command("nmcli -w #{wait} connection up primary", false)
+  command("nm-online -s -q --timeout=#{wait}")
+  # wait for IPv4, generate SSL self-signed cert and start proxy
   command("systemctl start foreman-proxy")
-  result
```
ares: we used to return result of a nm-online command which makes sense to me for a method called `configure_network`, shouldn't we store the result on line 88 and keep the result here?

lzap: Let me rewrite it and take both proxy and up command results into effect. Will do build for you shortly.

lzap:
To test:

* Boot the image in PXE-less, select DHCP mode and it should discover the host afterwards.
* Boot the image with fdi.dhcp_timeout=300 and do the same, it should wait 300 seconds instead of 120 default value.

Build: http://ci.theforeman.org/job/packaging_discovery_image/132/

The ISO should appear in half an hour at http://downloads.theforeman.org/discovery/scratch/lzap

Ping me if you run into issues or when you merge, I need to quickly do FDI scratch builds tomorrow.

ares:
@lzap the jenkins job seems to be red and http://downloads.theforeman.org/discovery/scratch/lzap is 404

mmoll:
let me edit the job (download.tf.org is on the CDM now, so the rsync doesn't work) and see if it's better then...

mmoll:
the job ran successful and the files are there: https://downloads.theforeman.org/discovery/scratch/lzap/

lzap:
Thanks a bunch @mmoll ! @ares ignore the version number, I forgot to push tags into my repo, the builds are okay it just has incorrect tag/name.

lzap:
I pushed this change:

```diff
diff --git a/root/usr/bin/generate-proxy-cert b/root/usr/bin/generate-proxy-cert
index 65383e8..b4f0e89 100755
--- a/root/usr/bin/generate-proxy-cert
+++ b/root/usr/bin/generate-proxy-cert
@@ -10,7 +10,7 @@ exportKCL
 
 DIR=/etc/foreman-proxy
 DAYS=${KCL_FDI_PROXY_CERT_DAYS:-999}
-WAIT=${KCL_FDI_NMWAIT:-120}
+WAIT=${KCL_FDI_IPWAIT:-120}
 SECONDS=0
 while (( SECONDS < $WAIT )); do
   sleep 1
```

Users can now fine-tune all timeouts:

* fdi.dhcp_timeout - DHCP NetworkManager timeout in seconds (300 by default)
* fdi.ipwait - wait time for IP to be available in proxy SSL cert start (120 by default)
* fdi.nmwait - nmcli --wait option for NetworkManager (120 by default)

https://github.com/theforeman/theforeman.org/pull/1000

lzap:
New build running at: http://ci.theforeman.org/job/packaging_discovery_image/136/

ares:
Thanks @lzap, merging!

lzap:
Thanks building 3.4.3 upstream release.

PR #101, opened 2018-01-23
Fixes #18715 - support for fdi.script option

lzap:
New `fdi.script` option is added, it requies base64-encoded gzipped script which is executed during boot. In PXE-less mode (when BOOTIF is not present on kernel commandline) it is executed before NetworkManager is started, in PXE mode it is executed after NetworkManager finished configuration and IP is assigned.

To generate compatible string use:

    echo "echo 'This is a test'" | gzip -9 | base64 -w0
    H4sIAK4hZ1oCA0tNzshXUA/JyCxWAKJEhZLU4hJ1LgA92U9qFgAAAA==

and then append the option to kernel command line. See system journal for output (grep for `discovery-script`).

    APPEND ... fdi.script=H4sIAK4hZ1oCA0tNzshXUA/JyCxWAKJEhZLU4hJ1LgA92U9qFgAAAA==

### DIFF
--- a/20-packages.ks
+++ b/20-packages.ks
@@ -25,6 +25,7 @@ facter
 ethtool
 net-tools
 dmidecode
+bind-utils
 virt-what
 
 # Foreman proxy
--- a/root/usr/bin/discovery-script
+++ b/root/usr/bin/discovery-script
@@ -1,0 +1,11 @@
+#!/bin/bash
+source /etc/default/discovery
+source /usr/share/fdi/commonfunc.sh
+exportKCL
+[ -d /tmp/facts ] || mkdir /tmp/facts
+defaultscript=${KCL_FDI_SCRIPT:-H4sIAOj1ZloCA0tNzshXUPLLVyhOLsosKFEoSCwuTk1R4gIAIbwTwhgAAAA=}
+basescript=${1:-$defaultscript}
+cat /usr/share/fdi/script-helpers.sh > /tmp/discovery-script
+echo "$basescript" | base64 -d | gunzip >> /tmp/discovery-script
+[[ $? -eq 0 ]] && ( bash /tmp/discovery-script | tee /tmp/discovery-script.log )
+rm /tmp/discovery-script
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -39,6 +39,8 @@ systemctl enable foreman-proxy.service
 systemctl enable discovery-fetch-extensions.path
 systemctl enable discovery-start-extensions.service
 systemctl enable discovery-menu.service
+systemctl enable discovery-script-pxe.service
+systemctl enable discovery-script-pxeless.service
 
 # register service is started manually from discovery-menu
 systemctl disable discovery-register.service
@@ -106,7 +108,7 @@ echo " * setting up journald and ttys"
 systemctl disable getty@tty1.service getty@tty2.service
 systemctl mask getty@tty1.service getty@tty2.service
 echo "Storage=volatile" >> /etc/systemd/journald.conf
-echo "RuntimeMaxUse=15M" >> /etc/systemd/journald.conf
+echo "RuntimeMaxUse=25M" >> /etc/systemd/journald.conf
 echo "ForwardToSyslog=no" >> /etc/systemd/journald.conf
 echo "ForwardToConsole=no" >> /etc/systemd/journald.conf
 systemctl enable journalctl.service
--- a/README.md
+++ b/README.md
@@ -81,11 +81,21 @@ http://theforeman.org/plugins/foreman_discovery/
 Building
 --------
 
+A host with either Fedora or CentOS 7 is required. RHEL 7 cannot be used as
+it is missing core dependency (livecd-tools), but this can be workarounded
+by installing it from CentOS 7 repositories (and two dependencies). Grub2
+EFI and Shim packages are only required if the resulting ISO must boot on
+UEFI systems.
+
 Install the required packages:
 
 ```
-$ sudo yum install livecd-tools pykickstart isomd5sum
+$ sudo yum install livecd-tools pykickstart isomd5sum syslinux \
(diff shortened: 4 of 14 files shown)


20-packages.ks
```diff
@@ -66,9 +66,13 @@ unzip
 binutils
 
 # For UEFI/Secureboot support
-grub2-efi
 efibootmgr
+grub2-efi
 shim
+# These were renamed and split in EL 7.4+
+grub2-efi-x64
+grub2-efi-x64-cdboot
```
lzap: This line is actually the one which fixes missing EFI loader. All the rest is just documentation around building or the script feature.

lzap:
Here is a scratchbuild, won't last long, hurry up: http://downloads.theforeman.org/discovery/scratch/lzap/

root/usr/bin/discovery-script
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+basescript=${KCL_FDI_SCRIPT:-H4sIAOj1ZloCA0tNzshXUPLLVyhOLsosKFEoSCwuTk1R4gIAIbwTwhgAAAA=}
+echo "$basescript" | base64 -d | gunzip > /tmp/discovery-script
```
stbenjam: I think there should be a little more error checking here.  COMMAND_LINE_SIZE isn't very big: 2K on my Fedora x86 box.  A script end up truncated, and I know it's an extreme and contrived example but "rm -rf /tmp/foo" could end up as "rm -rf /":

```
echo "rm -rf /tmp/foo" | gzip -9 | base64 -w0 | cut -c 1-26 | base64 -d | gunzip
```

Which I know isn't a big deal on the discovery image, but base64 will return an error code if the input isn't quite valid, can we check it?

stbenjam:
This is pretty clever!

root/usr/bin/discovery-script
lzap: Ok it looks like if a single character is missing (except `=` which is safe to cut) gunzip return non-zero code. I will take advantage and fail.

lzap:
Fixed, I added also a helper script which will encode file and show total length and the limit:

```
[lzap@box foreman-discovery-image]$ cat 20-packages.ks | aux/discovery-script-creator
fdi.script=H4sIANzQaVoCA3VV224cNwx911cIMAIkTeVtjKQNWhR9qdMYSJogbp4NjcSZIVY3iNRe8vWlZsZ2ErT74LUkkuI5PEf7pFi3txOQNgZOLjQPPjtSg6VZ7aEmCGrATB4OyUZQU23D1frXcM6BFFyNVGqeSBVLdPSq5IDu7HKFxigBbt67nEacVM2ZRwxA6tyiOmA0ERNGG5R1Bb3ikYsKwRfrFVKO/hVJnLrQf1brGuuIRJgm7aGQfjozF/p1txva9AVDsJcV/Gz50uW4ozkf7+Tg0k34B/rfX1y9fvX6xU/PFNuqJLz0qm+sY6hqXL+A5w5IJeANmY/owWUP0mplc5TiS5ogizZpAX06K2o+qw8F0s3H9zcP/2wVsiyJghrXFLOmCHnnCaKhaCvfLXt3HsnlA9TznfAxQb/nJklb0hseQD8cy0xO4Lby94USHPlhMVriuwmY4cRqP/iFPxAupk4dtVJyZSVDINVnodgV3+JCyIcUzjoBiAb0AM42Ah3keufXC3VtifSuUd0NmHYjVjjaEEwex4AJjIteqrQk4/bImJMcnn/TxbKb5eYhonTldSvEFWy81J8gCiqdkwNdIYAlORaYJLlS6Yg8a2SNpO3BYrCD9Ht/6wLs9vatts51MCvXs3EBIfHjmqBKxSVYhs9SKgSNnVsZvOjeNs7RMrrerJZJ6Ubqb+Bjrvv3NskwqmoNl+s+9/44CwJvGXTXhrbcaWEUc7S0Kes69Va1wMRShHaxUFrNsOpHf75+c7O7BdcqDGKKh7HAiH0dp7p5TDYUzRgl758ZZB5HqJ2rbkavbfKaSugUJX39Tv9y+fL5Y6I5/fzy25VxvpdfKi6nUnadLIu4Naxt8wx6UaFUJRZaoOoSmuhHLF5lgip666M63Rv/8QkYhIArRWfirm/OVaqYuLGoLv7zIz18vH+EhNtNExf/Fy3hnwT6W+n3WqZYKgor74TfU5cYOaG8S2/BpczXW181ov+CBBXdwuARu/aIViqUuBREzntJBhb1oLxPRkAbUcq8vWTmPmdz4tLU0ncnT1ynBxQNmu4/IzaqsT8jEtUtp1d+RB5tHJXxsVqRlwmHeNVDqN8uaHplN1uZbB71d4/qj+LLgWAFpJ8uizVLvLI5mOf8TMB89xobN4Pbr7sd4ZK1Rf2gFpmJYKnb7phb6A+BTij+lNGEnAmUGeUnoloT8pTpYbXZ16TM8r4/geTVv5JFDdNcBgAA
Maximum length: 512 (minus what's on kernel command line)
Actual length: 1243
Consider shell minifier or stage2 script downloaded from network.

[lzap@box foreman-discovery-image]$ echo "Test" | aux/discovery-script-creator
fdi.script=H4sIACHRaVoCAwtJLS7hAgDCGnj6BQAAAA==
Maximum length: 512 (minus what's on kernel command line)
Actual length: 47
```

stbenjam:
This is pretty sweet, I just built an image off the latest changes and it works perfectly.  The helper functions are nice, especially since the kernel cmdline is so limited. 

ACK from me

lzap:
Thanks, note for googlers - the example in the repo (configuring bonding) does send facts correctly, but Discovery currently errors out with "No IP address detected". This is because in the current design we need all interfaces to have IP address and MAC address, which is not the case for bonds. This is a RFE: http://projects.theforeman.org/issues/18583

PR #107, opened 2018-07-25
Fixes #24378 - added Hyper-V drivers

lzap:
Don't test this with real Hyper-V hypervisor if you don't have any. Just build new image and then verify that before and after the patch the drivers are now present:

    lsinitrd /home/libvirt/fdi-image/initrd0.img | grep hyper


lzap:
@ares this would be easy merge, needed for downstream FDI build

ares:
Thanks @lzap, merging!

PR #110, opened 2018-09-17
Fixes #24956 - remaster uses guestfs for extraction

lzap:
Previously we used `mount` to extract the original ISO, but this required `root`. Since we want to use remaster script during RPM package `foreman-discovery-image` build to put correct kernel options, this is not possible in buildroot/koji environment.For this reason I changed the extraction to `isoinfo`, the only available tool in RHEL7. This worked in Fedora, but older version of `isoinfo` had problems with Joliet extensions and this was leading to unextracted files. Commit 0b88c3770420684d698f3a1783ffcc8103069853.

I found a different solution - to use `guestfs` tools available both in Fedora and RHEL7. It allows mounting of VM images and ISO files, it's slightly slower but this is fully maintained package by Red Hat engineering and it does the job correctly. We simply cannot use `7zip` to extract ISO file as it is not available, so this is the only option.

The other option was to use `fuseiso` but this tool has been deprecated and removed from Fedora, so we'd need to solve this problem in near future.

### DIFF
--- a/aux/remaster/discovery-remaster
+++ b/aux/remaster/discovery-remaster
@@ -5,12 +5,19 @@ if [ -z "$2" ]; then
   exit 2
 fi
 
-which isoinfo dd mkisofs isohybrid implantisomd5 >/dev/null || ( echo "Command(s) missing, install required tools" && exit 3 )
+REQ_CMDS="guestmount dd mkisofs isohybrid implantisomd5"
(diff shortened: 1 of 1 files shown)


ares:
looks reasonable :+1:

lzap:
Thanks all we are all set, buildroot fixed.

PR #111, opened 2018-09-17
Refs #24956 - remaster uses guestfs without fuse

lzap:
It turned out that `fuse` kernel module is not enabled in Fedora/RHEL buildroots. This avoids fuse completely.


lzap:
Thanks, it's a tool for manipulating VM or ISO images.

PR #113, opened 2019-07-12
Copy remastered grub.cfg into efiboot.img to fix UEFI discovery

0x41414141:
This should fix remastered images not auto provisioning in EUFI boot environments. 
mkisofs was being called with "-e isolinux/efiboot.img" which contained the EUFI partition and grub.cfg which was not being updated with the kernel command line options when "$TMP_NEW/EFI/BOOT/grub.cfg" was being generated.
This patch will mount the efiboot.img, copy in the newly updated grub.cfg, unmount and unlink the temporary directory.

### DIFF
--- a/aux/remaster/discovery-remaster
+++ b/aux/remaster/discovery-remaster
@@ -83,6 +83,12 @@ menuentry 'Check media' --class fedora --class gnu-linux --class gnu --class os 
 }
 EOGR
 
+TMP_NEW_MNT=$(mktemp -d)
+mount -o loop $TMP_NEW/isolinux/efiboot.img $TMP_NEW_MNT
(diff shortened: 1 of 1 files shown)


lzap:
Hello, sorry for the delay. What was the error symptoms?

lzap:
I just remastered an ISO and it boots fine on libvirt UEFI mode, however I am willing to test and accept the patch if you explain me the symptoms.

lzap:
Well it boots fine on libvirt, merging. Thanks!

0x41414141:
Sorry for not explaining further - in UEFI mode the discovery wouldn’t automatically work on the boot image - under legacy boot it would find the boot nic and discover foreman server find though - seemed to be due to missing kernel options in the efiboot image. Thanks for merging

PR #123, opened 2020-05-23
Added more options to customize bonding interface (device, bootproto).

johnnys66:
Use case:
- no provisioning network or DHCP available (PXE not an option)
- using a customized boot image with static network config (iPXE boot image) from where FDI is booted with all fdi.pxXXX  required parameters and custom facts 
- fdi.script option is used

Requirement:
- cfg_bond needs to configure static address and allow to use for example a bonding interface with `DEVICE=bond0` and `NAME=primary`, given that an NM connection named `primary` is a requirement for `generate-proxy-cert` to be successful.

### DIFF
--- a/root/usr/share/fdi/script-helpers.sh
+++ b/root/usr/share/fdi/script-helpers.sh
@@ -30,13 +30,13 @@ hwaddr() {
 }
 
 cfg_eth() {
-  NAME=${1:-eth0}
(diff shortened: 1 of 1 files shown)


johnnys66:
cfg_eth() also needs to be able to have `NAME` different from `DEVICE` in the above use case, in situations where no bonding will be used.
Also fixes wrong variable name `$DEV` in function cfg_eth().

PR #125, opened 2020-08-19
Fixes #30662 - do not ignore IPv6

lzap:
Radovan is testing this one for me. Can you take a look @adamruzicka?


lzap:
Context: IPv6 was disabled until now, this patch changes "ignore" to "auto" in network manager. But still it allows to use `fdi.ipv6.method` kernel option to override this value in case this hits existing IPv4 customers.

root/usr/bin/nm-configure
```diff
@@ -49,13 +51,13 @@ autoconnect-priority=1
 [802-3-ethernet]
 mac-address=$mac
 [ipv4]
-method=manual
+method=$ipv4_method
 address1=$ip
 gateway=$gw
```
adamruzicka, line 56: Since we're changing the default from manual to auto, how does networkmanager behave when we pass address and gateway, but at the same time set method to auto?

lzap: Good point, this is an overlook. This needs to be manual.

It brings another missed feature - IPv6 is not yet done for manual mode (when you attach the ISO and use TUI to configure IP). So it is impossible to reach to this state yet.

lzap: Corrected, however as I said this command is called from the text user interface and the screen does not yet have elements to work with.

root/usr/bin/nm-configure
```diff
@@ -55,7 +57,11 @@ gateway=$gw
 dns=$dns;
 dhcp-send-hostname=$sendhost
 [ipv6]
-method=ignore
+method=manual
+address1=$ip
+gateway=$gw
```
adamruzicka, line 62: If I'm reading this right this way the same value is used for address and gateway for both ipv4 and ipv6. Not sure if network manager is able to gracefully deal with ipv4 address being given as ipv6 address and vice versa, depending on which address the user inputs.

lzap: Right, I think the most reasonable solution is to introduce new command `primary-static6`.

adamruzicka: Sounds good to me

lzap: How about this?

adamruzicka (commented):
Looks good to me, can we get Radovan's blessing before we merge it?

lzap:
> Looks good to me, can we get Radovan's blessing before we merge it?

Absolutely. I am actually working on IPv6 support for the TUI, it is trivial patch. Testing it currently.

lzap:
Do not merge, I am looking into TUI, cert generation issues as well as IPv6 facts error on discovery side. Many issues :-(

lzap:
http://people.redhat.com/~lzapleta/temp/rdrazny/

This build should work when booted locally (use direct kernel command line in libvirt) or via PXE. Still seeing an error after facts are uploaded tho, this is a bug in discovery.

lzap:
I think we can merge because I will be doing big refactoring in order to upgrade to Facter 4.0. There will be another PR which will need a lot of QA attention.

lzap:
I think we are good, at least to have an initial support for IPv6 now. This is not tested end-to-end but IPv6 was not working. Let's merge so we can put a build for 2.3 out.

PR #127, opened 2020-08-25
Minimize image even more

lzap:
The FDI system is base RHEL7 with a minimum set of packages, because
of slow and unreliable PXE the goal is to keep the image size as small
as possible, therefore some files are completely removed
(documentation, man pages, i18n). When we started with this, the image
size was reasonable (145MB) but then it was only growing. Today it's
415MB, it's growing quite fast. Although we added SCL Ruby for some
Foreman services, that does not justify such growth.

Thus this patch. Overall I was able to bring down the size from 415M to 335M
with these changes. Not bad, I could go even further analyzing one package
after another, but we plan to remove Ruby from the image so I will rather
invest time into this.

### DIFF
--- a/20-packages.ks
+++ b/20-packages.ks
@@ -83,19 +83,44 @@ e2fsprogs
 bzip2
 system-storage-manager
 
+# finding duplicite files saves about 1% of the compressed size
+rdfind
+
 ######################
 # Packages to Remove #
 ######################
+#
+# Some ideas from:
+#
+# https://github.com/weldr/lorax/blob/rhel7-branch/share/runtime-cleanup.tmpl
 
 # Red Hat Enteprise Linux subscription tool
 -subscription-manager
 
-# Generic and wireless tools
+# Generic and wireless tools and firmware
 -prelink
 -setserial
 -ed
 -authconfig
 -wireless-tools
+-iwl7260-firmware
+-iwl3160-firmware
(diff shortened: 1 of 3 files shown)


20-packages.ks
```diff
@@ -105,10 +128,22 @@ system-storage-manager
 -dmraid
 -lvm2
 
+# sound and video
+alsa-lib
+alsa-firmware
+alsa-tools-firmware
+ivtv-firmware
```
adamruzicka, line 135: Wait, are you adding those?

adamruzicka (commented):
Overall looks sensible, apart from the sound and video part

20-packages.ks
lzap: Typo, thanks for saving another 3MB of compressed size.

lzap:
Rebased, RTM.

FYI there is also a patch for livecd-tools that increases compression ratio at the cost of more memory usage which saves huge chunk as well: https://github.com/livecd-tools/livecd-tools/pull/163

adamruzicka (approved):
Looks good to me

lzap:
According to discussion in pagure.io thread I am also removing liquidio and netronome firmware:

```
+-liquidio-firmware
+-netronome-firmware
```

Tested. Merging.

PR #129, opened 2020-09-16
Add --flat-squashfs option to livecd-creator

lzap:
This shaves whooping 50MB!

### DIFF
--- a/build-livecd-root
+++ b/build-livecd-root
@@ -35,7 +35,7 @@ cd $tmpdir
 echo Working in directory $tmpdir
 
 echo "* Running livecd-creator"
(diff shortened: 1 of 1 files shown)


lzap:
Do not merge, the image unfortunately does not boot. Further discussion: https://pagure.io/releng/issue/9689

Need to take a look, but saving 50MB is really interesting.

lzap:
This needs a patch in Dracut for RHEL7, this is probably too late. But once we upgrade to RHEL8 we should use this option to keep the size down.

https://github.com/livecd-tools/livecd-tools/pull/109

PR #131, opened 2020-10-15
Use xz1m compression type when available

lzap:
Livecd-creator was extended with this new option. When it gets into Fedora the image will be slightly smaller.


lzap:
Need to wait for Fedora 34 until this feature is in livecd-creator.

lzap:
We now use lorax for EL8+ which supports arbitrary options. Closing.

PR #133, opened 2020-11-24
Use TFM SCL Ruby

lzap:
Newt library is no longer provided for system Ruby 2.0 so discovery menu does not show up. This fixes it.


lzap:
So far so good, I briefly tested this patch. Need to merge this so I can rebase more important changes on top of this.

PR #134, opened 2020-12-09
Fixes #31481 - fix nm configuration IPv6 ignore and remove facter2

lzap:
A typo in the config file, IPv4 requires "disabled" but IPv6 requires "ignore". In Fedora man pages this is both "disabled", it recently changed upstream likely.

The second patch finalizes removing facter 2. FDI is now fully on SCL Ruby and Facter 4.

### DIFF
--- a/root/usr/share/fdi/facts/discovery-facts.rb
+++ b/root/usr/share/fdi/facts/discovery-facts.rb
@@ -18,41 +18,36 @@
 # also available at http://www.gnu.org/copyleft/gpl.html.
 
 require 'resolv'
-require 'facter/util/ip'
-require '/usr/lib64/ruby/vendor_ruby/discovery.rb'
+if File.exist? '/opt/rh/rh-ruby25/root/usr/share/ruby/vendor_ruby/discovery'
+  require '/opt/rh/rh-ruby25/root/usr/share/ruby/vendor_ruby/discovery'
+else
+  require_relative '../../../../opt/rh/rh-ruby25/root/usr/share/ruby/vendor_ruby/discovery.rb'
+end
 
 def discovery_bootif
   # PXELinux dash-separated hexadecimal *without* the leading hardware type
   cmdline('BOOTIF', cmdline('fdi.pxmac', detect_first_nic_with_link)).gsub(/^[a-fA-F0-9]+-/, '').gsub('-', ':') rescue '00:00:00:00:00:00'
 end
 
-Facter.add("discovery_version") do
-  setcode do
-    File.open('/usr/share/fdi/VERSION') {|f| f.readline.chomp}
+if File.exist? '/usr/share/fdi/VERSION'
+  Facter.add("discovery_version") do
+    setcode do
+      File.open('/usr/share/fdi/VERSION') {|f| f.readline.chomp}
+    end
   end
 end
 
-Facter.add("discovery_release") do
-  setcode do
-    File.open('/usr/share/fdi/RELEASE') {|f| f.readline.chomp}
+if File.exist? '/usr/share/fdi/RELEASE'
+  Facter.add("discovery_release") do
+    setcode do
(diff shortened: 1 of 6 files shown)


root/usr/share/fdi/facts/discovery-facts.rb
```diff
@@ -44,18 +51,6 @@ def discovery_bootif
...
-    result = Facter.fact("ipaddress").value
-    required = discovery_bootif
-    Facter::Util::IP.get_interfaces.each do |iface|
-      mac = Facter::Util::IP.get_interface_value(iface, "macaddress")
-      result = Facter::Util::IP.get_interface_value(iface, "ipaddress") if mac == required
-    end
-    result
-  end
-end
-
```
lzap, line 58: This fact was never used for anything.

root/usr/share/fdi/facts/discovery-facts.rb
```diff
@@ -71,45 +66,6 @@ def discovery_bootif
...
-      Facter.add("#{fact}_#{Facter::Util::IP.alphafy(interface)}", :timeout => 10) do
-        confine :kernel => "Linux"
-        setcode do
-          value
-        end
-      end
-    end
-  end
-end
-
```
lzap, line 112: I am dropping this for now, we can add later.

root/usr/share/fdi/facts/openlldp.rb
```diff
@@ -82,7 +79,7 @@ def grep_v(cond)
             when 'chassisID'
               output.split("\n").each do |line|
                 ether = $1 if line.match(/MAC:\s+(.*)/)
-                result = Facter::Util::Macaddress.standardize(ether)
+                result = ether&.tr('-', ':')&.downcase
               end
```
lzap, line 83: Facter 4 no longer provides the Util class, so let's find list of devices a different way.

root/usr/share/fdi/facts/discovery-facts.rb
```diff
@@ -18,23 +18,30 @@
 # also available at http://www.gnu.org/copyleft/gpl.html.
 
 require 'resolv'
-require 'facter/util/ip'
-require '/usr/lib64/ruby/vendor_ruby/discovery.rb'
+if File.exist? '/opt/rh/rh-ruby25/root/usr/share/ruby/vendor_ruby/discovery'
+  require '/opt/rh/rh-ruby25/root/usr/share/ruby/vendor_ruby/discovery'
+else
+  require_relative '../../../../opt/rh/rh-ruby25/root/usr/share/ruby/vendor_ruby/discovery.rb'
```
adamruzicka, line 24: This feels odd, what would be the absolute variant of this?

root/usr/share/fdi/facts/openlldp.rb
```diff
@@ -66,7 +63,7 @@ def grep_v(cond)
   }
 
   # Remove interfaces that pollute the list (like lo and bond0).
-  Facter.value('interfaces').split(/,/).grep_v(/^lo$|^bond[0-9]/).each do |interface|
+  Dir.glob("/sys/class/net/*").map{|x| Pathname.new(x).basename.to_s}.grep_v(/^lo$|^bond[0-9]/).each do |interface|
```
adamruzicka, line 66: Any reason for `Pathname.new(x).basename.to_s` over `File.basename(x)`?

adamruzicka (commented):
Two nitpicks inline, otherwise lgtm

root/usr/share/fdi/facts/discovery-facts.rb
lzap: This is only used when running this from git, no big deal. It would require users to modify LOAD_PATH otherwise.

root/usr/share/fdi/facts/openlldp.rb
lzap: Fixed thanks.

PR #135, opened 2020-12-22
Remove references to discovery_bootip

petski:
discovery_bootip was removed in 68f2446aa90c5436e0ca85925fd9b6ea222fdb8e


lzap (approved):
Yes, thanks. I realized it was never used in the plugin itself. Feel free to put it back if you need it.

PR #140, opened 2021-11-01
Switch to using CentOS 8

srwalter:


build-livecd-root
```diff
@@ -1,95 +1,10 @@
...
-    fi
-}
-
-if [ $(id -u) != 0 ]; then
-  echo "Please run this script as root"
-  exit 1
+rm /var/tmp/lmc-work-* -rf
+if [ ! -e CentOS-8.4.2105-x86_64-dvd1.iso ]; then
+    wget http://ftp.osuosl.org/pub/centos/8.4.2105/isos/x86_64/CentOS-8.4.2105-x86_64-dvd1.iso
 fi
```
lzap, line 6: Allright, this is a big change. Previously the script was building from live repos, this change requires the full DVD and then only installs minimal system from it. On our CI this will require more space.

Not that I dislike it, I am just trying to summarize what this can possibly affect. Another thing to consider is 8.4 (or 8.5 not sure) is the last CentOS 8 release, so do we want to use CentOS Stream? It appears to have a isodirect redirector to the "latest" DVD:

wget http://isoredirect.centos.org/centos/8-stream/isos/x86_64/CentOS-Stream-8-x86_64-latest-dvd1.iso

The problem with that is that it changes quite often, I currently see a build that is one day old.

My suggestion:

* name the ISO as `install.iso` or something (can be a symlink too)
* let's parametrize the ISO URL
* let's use http://isoredirect.centos.org/centos/8-stream/isos/x86_64/CentOS-Stream-8-x86_64-latest-dvd1.iso as the default value
* for development builds this is fine
* for production builds we will probably keep a copy of DVD and build from that so we can repeat builds (unfortunately CentOS Stream project deletes all old ISOs)
* users can opt-in to building from RHEL, AlmaLinux or other alternatives

What you think?

build-livecd-root
```diff
@@ -1,95 +1,10 @@
...
-
-which livecd-creator isohybrid implantisomd5 >/dev/null || ( echo "Command(s) missing, install required tools" && exit 2 )
```
lzap, line 22: Can you keep this dirty test for missing tools? Obviously check only the relevant ones.

20-packages.ks
```diff
@@ -11,29 +16,23 @@ chkconfig
 rootfiles
 yum
 vim-minimal
-acpid
-tftp
```
lzap, line 15: TFTP utility is needed to download extensions, is that not available anymore in EL8?

20-packages.ks
```diff
@@ -11,29 +16,23 @@ chkconfig
...
 lldpad
-isomd5sum
```
lzap, line 17: I think this way needed so ISO checksum can be called from the bootloader (check iso and boot). Not sure about it, but I think it goes into initramdisk when it is built, then it is not needed after boot of course.

25-minimize.ks
```diff
@@ -68,9 +68,6 @@ for log in yum.log dracut.log lastlog yum.log; do
     truncate -c -s 0 /var/log/${log}
 done
 
-echo " * removing /boot, since that lives on the ISO side"
-rm -rf /boot*
-
```
lzap, line 73: IIRC this was breaking something in EL8.

build-livecd-root
```diff
@@ -1,95 +1,10 @@
...
-EOF
-
-sha256sum initrd0.img vmlinuz0 > SHA256SUM
-
-cd ..
-tarball=$destdir/fdi-image-${last_tag#release-}.tar
-chmod 666 fdi-image/*
-tar cf $tarball fdi-image/
-ls -lh $tarball
-chown --reference $srcdir/build-livecd $tarball
```
lzap, line 90: We need all of this to remaster (change kernel options), make the iso bootable via USB sticks and ship PXE images.

lzap (changes requested):
Ok thanks, this was on my TODO for January 2022 but since you started, let's keep pushing this.

Is this your own work of have you followed https://community.theforeman.org/t/building-discovery-image-on-el8/23719 ?

I will try this now, I need to update the build script tho - we need to build:

* the ISO for booting locally
* the PXE files
* make sure it boots from USB stick too
* I would like to keep some script functionality

Let me test this first. Don't do any adjustments for now please.

lzap:
For the record, it looks like all packages are being downloaded from network rather than from the DVD

```
11:27:08,277 INFO anaconda:librepo: Downloading: http://ftp.fi.muni.cz/pub/linux/centos/8.5.2111/BaseOS/x86_64/os/Packages/virt-what-1.18-12.el8.x86_64.rpm
11:27:08,303 INFO anaconda:librepo: Downloading: http://ftp.fi.muni.cz/pub/linux/centos/8.5.2111/BaseOS/x86_64/os/Packages/which-2.21-16.el8.x86_64.rpm
11:27:08,396 INFO anaconda:librepo: Downloading: http://ftp.fi.muni.cz/pub/linux/centos/8.5.2111/BaseOS/x86_64/os/Packages/xfsprogs-5.0.0-9.el8.x86_64.rpm
1
```

I think we should use netboot instead since downloading a 10GB DVD has no effect. We could update the kickstart to use DVD tho.

I am currently hitting this bug when building CentOS 8 Stream: https://bugzilla.redhat.com/show_bug.cgi?id=1955836

lzap:
I hit two bugs, but I was able to hotfix Anaconda and figure out everything. I have an ISO ready, need to test it, then do PXE images building and update docs.

lzap:
WIP: https://github.com/theforeman/foreman-discovery-image/pull/141

lzap:
https://community.theforeman.org/t/foreman-discovery-image-3-8-el8-alpha/26619

evgeni:
given #141 is merged, can this be closed?

lzap:
Yes thank you I kept your commit and stacked bunch of mine on top of that.

We are very close to the release, I am finalizing testing of the latest build, thanks @srwalter !

PR #142, opened 2021-12-23
Minimize image size after EL8 upgrade

lzap:
So FDI grew to 480MB:

```
-rw-r--r--. 1 lzap lzap 483M 23. pro 16.06 fdi-3.8.0-d737a06.iso
```

This PR should improve this. WIP, do not merge.


lzap:
Interesting, the ISO has the very same size after this patch. It either does not work or lorax already use xz with 1MB dict size. Need to take a look later.

PR #147, opened 2022-05-05
Add new remaster script compatible with RHEL8 based FDI 4

pbburke:
Only changes are to linuxefi and initrdefi lines


pbburke:
Sorry, this tested fine in a VM but has some further issues I'm going to look into

PR #148, opened 2022-05-24
Module is required now for rubygem-facter to install

lzap:
This fixed the problem reported in

https://community.theforeman.org/t/unable-to-build-and-or-remaster-fdi-4-0-3/28678/4

Scratchbuild: https://ci.theforeman.org/job/foreman-discovery-image-release/50/


lzap:
I think our CI does not support repo_owner anymore, it used to handle it in the past. So the job failed, but it works locally.

lzap:
Building new nightly build: https://ci.theforeman.org/job/foreman-discovery-image-release/51/

PR #150, opened 2022-06-22
Fixes bug #34914 - update remaster script for RHEL8-based 4.0 FDI 

pbburke:
Fixes Bug #[34914](https://projects.theforeman.org/issues/34914)

### DIFF
--- a/aux/remaster/discovery-remaster
+++ b/aux/remaster/discovery-remaster
@@ -74,24 +74,24 @@ set gfxpayload=text
 set timeout=3
 search --no-floppy --set=root -l 'fdi'
 menuentry 'Discovery' --class fedora --class gnu-linux --class gnu --class os {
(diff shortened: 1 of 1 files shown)


lzap:
I need to setup an EFI VM to see this, can you explain "But the remaster script still turns the FDI ISO into a flat image with no partition structure.". I don't understand.

The remaster script touches nothing but config files, we actually ship kernel/initramdisk twice on the ISO, we should probably drop one pair to save about 50MB.

lzap:
Oh wait the original ISO is created with `-cache-inodes` option which actually makes a ISO hardlink to the kernel and initramdisk. This explains why remastered ISO is larger by 50MB. I will fix this.

Also I think I know what you mean now, with EL8 (lorax) the EFI ESP is now under `/images/` instead of `/isolinux` where the script expects it. Then it basically fallsback to BIOS only ISO not providing the correct option.

lzap:
Ok figured it out all, thanks for the report and the patch. It was more complicated than that. I fixed it in this commit: https://github.com/theforeman/foreman-discovery-image/commit/b94bf01e26be01edffc1290b8c726f2d6ecdceec

Please test now. It should be all good now, on top of that the remastered image is now 50MB smaller :)

Cheers.

PR #152, opened 2022-08-15
disable synced folders

evgeni:
rsync is currently broken in CentOS 8 Stream, but we don't even need it to be working as we do a fresh git clone inside the VM anyways


lzap (approved):
Right we do ssh the data out and in, good point.

PR #153, opened 2023-02-16
Remaster remove 0 from vmlinuz and initrd

stejskalleos:
Fix issue with not working iso after `remaster` command is used.


lzap (commented):
Interesting, it always was named like this. Nice.

shubhamsg199:
Hello,
I tested this patch on the satellite and I was able to provision with BIOS as well as UEFI.
Thanks @stejskalleos !

Ron-Lavi:
@stejskalleos do we need another review for it? cc @shubhamsg199

PR #154, opened 2023-03-24
fix: add ipmitool as dependency

petski:
It's used in discovery-facts.rb, and not installed without this change


lzap (approved):
Thank you, @stejskalleos is not working this week, so it might take a while to get in but I have no objections, this will not grow the size of the image too much.

PR #155, opened 2023-03-29
Describe alternative FDI PXE-boot approach

jloeser:
We notice more and more issues with PXE-boot/GRUB2 when loading initrd:

https://community.theforeman.org/t/error-cant-allocate-initrd/27322
https://community.theforeman.org/t/discovery-4-1-0-fails-with-cant-allocate-initrd/32558

This approach was successfully tested with FDI v4.1.0 on various setups.


lzap (approved):
Thanks!

PR #156, opened 2023-04-06
Disable systemd-journal-flush.service

jloeser:
This service tries to store volatile log data to persistent storage. As we don't have any persistant storage, it timeouts after 90s. We do not need this service.

Bug report: https://projects.theforeman.org/issues/36235

### DIFF
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -154,6 +154,9 @@ cat > /etc/udev/rules.d/83-enable-promiscuous-mode.rules <<'UDEV'
 ACTION=="add", SUBSYSTEM=="net", NAME!="lo", TAG+="systemd", ENV{SYSTEMD_WANTS}="enable-promiscuous-mode@%k.service"
 UDEV
 
+echo " * disable flushing log data"
+rm /usr/lib/systemd/system/sysinit.target.wants/systemd-journal-flush.service
+
 # extra modules for livecd-creator/livemedia-creator
 echo 'add_drivers+="mptbase mptscsih mptspi hv_storvsc hid_hyperv hv_netvsc hv_vmbus"' > /etc/dracut.conf.d/99-discovery.conf
 
--- a/22-discovery.ks
+++ b/22-discovery.ks
@@ -155,7 +155,7 @@ ACTION=="add", SUBSYSTEM=="net", NAME!="lo", TAG+="systemd", ENV{SYSTEMD_WANTS}=
 UDEV
 
 echo " * disable flushing log data"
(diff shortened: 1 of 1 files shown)


petski:
Out of curiosity: What's the reason to remove the unit file, instead of disabling it? In other parts of `22-discovery.ks`, a `systemctl disable ...`-approach is used, i.e. on line 13.

jloeser:
It does not remove the unit file but the symlink to that unit file to not start this service during boot.

Adding/removing symlinks of services to `*.wants` directories is how systemd works.

`systemctl enable/disable <service>` does exactly the same: creating/removing corresponding symlinks to that unit.

In scripts I prefer to work with symlink files directly as you might not have full systemd environment running.

lzap:
I sort of also lean towards using `systemctl` command instead. This gives me some SYS V init config files and symlinks memories and not all of these are good. But on more serious note:

You described exactly [how the enable command works](https://github.com/systemd/systemd/blob/main/src/systemctl/systemctl-enable.c), but then you say "might not have full environment running" which is contradictory. This command literally does the symlink removal, does not need anything to be running and it is shorter to write :-)

Anyway, I will let @stejskalleos the current maintainer to decide. He will be back in a bit.

jloeser:
> You described exactly [how the enable command works](https://github.com/systemd/systemd/blob/main/src/systemctl/systemctl-enable.c), but then you say "might not have full environment running" which is contradictory. This command literally does the symlink removal, does not need anything to be running and it is shorter to write :-)

That statement was not referring to this project. If you assemble or debug images it happens that there is no `systemctl` available. But yes, in this case we can go with `systemctl` commands to keep it consistent.

lzap:
Oh if it does not work then symlink is fine, or whatever you find appropriate. Thanks for testing, I was not expecting that not to work!

jloeser:
I think current version with `systemctl mask systemd-journal-flush.service` is best compromise.

stejskalleos:
Thanks @jloeser @lzap !

ekohl:
Late to the party, but this is https://bugzilla.redhat.com/show_bug.cgi?id=2174645.

lzap:
Might be, but livecd does not have any persistent storage, I mean, other than the overlay over readonly ISO. Makes sense to just straight up disable it.

ekohl:
Yes, in this case it's a good solution too but I wanted to link back the root cause which you can also see if you boot the latest vagrant image. systemd in CentOS Stream 8 just has a very annoying bug right now and there's little you can do in some cases. Luckily here the solution is actually a good thing we can keep even after the original bug is fixed.

PR #159, opened 2023-06-08
Fix links

kitgerrits:
Fixes 36488

I'm new to this and thought I would check out the tips for suggesting improvements.
The way the bottom of the README renders, it looks like someone swapped `[]` and `()`.
```
Please follow our (generic contributing guidelines for Foreman)[http://theforeman.org/contribute.html#SubmitPatches]. Make sure you create (an issue)[http://projects.theforeman.org/projects/discovery/issues/new] and select "Image" Category.
```

This minor change should fix that.


stejskalleos (changes requested):
Hi, can you squash the commits into the one? Also, I suggest changing the URL from http://theforeman.org/contribute.html#SubmitPatches to http://theforeman.org/contribute.html, the headlines change often so it's safer to not include them in the links.

kenyon:
This was fixed in #166.

PR #162, opened 2023-10-30
VLAN id entered by users on discovery-menu interface page never used properly

nixfu:
The ability for users to enter a VLAN in the discovery interface menu is currently broken.  If the user enters it, the value never gets used because the Network Manager config file is not proper format for the data.

The culprit is the nm-configure script trying to improperly insert a newline in a bash variable assignment.

This makes the /etc/NetworkManager/system-connections files look improperly like this:
[vlan]\nid=123

Instead of:
[vlan]
id=123

With this fix, adding the newline correctly, NetworkManager will now use the vlan id entered by the user on the discovery-menu interface selection page.

### DIFF
--- a/20-packages.ks
+++ b/20-packages.ks
@@ -79,6 +79,7 @@ mdadm
 xfsprogs
 e2fsprogs
 bzip2
+tcpdump
 
 ######################
 # Packages to Remove #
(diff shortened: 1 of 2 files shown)


nixfu:
This is a dupe.

PR #163, opened 2023-11-14
Do not remove wifi drivers and packages

stejskalleos:


stejskalleos:
With the wifi & network drivers, the final size of the image is double, which is not great TBH.

```console
-rw-r--r--. 1 root root 1021MB Nov 13 04:44 install.iso
```

sayan3296:
An ISO was created and shared by @stejskalleos with me. I forwarded that ISO to one of the affected end-user and after testing the ISO on a Dell Latitude 7560  laptop, the user confirmed that he no longer observes any issues with discovering the laptop or with the overall discovery process.

He is doing more testing but for now, this is a positive confirmation that I have got and hence shared here.

PR #168, opened 2024-09-05
Fix repo URLs

mschnepf:
The CentOS Stream repositories are no longer available at the current URLs since the CentOS Stream 8 end of builds was May 31, 2024; see [CentOS download page](https://www.centos.org/download/). However, a snapshot of the old repo, without updates, is available at https://vault.centos.org/.
Comment from vault.centos.org:
```
This is _NOT_ an updated tree for installing CentOS Linux : It is a snapshot of the older trees that have been removed from the main CentOS servers as new point releases are released.
This is provided for reference and to provide access to older archived versions, and we do not put security updates into the trees on this server.
Please see http://www.centos.org/download for active versions of CentOS Linux
```


stejskalleos (approved):

stejskalleos:
Thanks @mschnepf

PR #171, opened 2025-02-17
IPv6 Custom facts and kexec

stejskalleos:
* Report custom discovery facts for IPv6
* Improve the custom_facts method
* Generate Rubocop todo (just for better dev experience)


stejskalleos:
[fdi-dev-5.0.1-49c8193.iso](http://fdi-el9.stejskalleos.cz/fdi-dev-5.0.1-49c8193.iso)
[fdi-image-dev-5.0.1-49c8193.tar](http://fdi-el9.stejskalleos.cz/fdi-image-dev-5.0.1-49c8193.tar)

root/usr/share/ruby/vendor_ruby/discovery/menu.rb
```diff
@@ -106,22 +106,29 @@ def perform_upload proxy_url, proxy_type, custom_facts
...
+def new_custom_facts(mac)
+  custom_facts = {}
+  custom_facts['discovery_bootif'] = mac
+  custom_facts['discovery_kexec'] = command('kexec --version')
+
+  ip_cidr4, gw4, dns4 = detect_ipv4_credentials('primary')
+  ip_cidr6, gw6, dns6 = detect_ipv6_credentials('primary')
+
+  ip4 = ip_cidr4.split('/').first if ip_cidr4
+  ip6 = ip_cidr6.split('/').first if ip_cidr6
```
stejskalleos, line 118: One thing is that this does not ignore local addresses like `fe80::2aa:bbff:fe17:1701`.

So if the host has IPv4 and local IPv6, the IPv6 will be prioritized in the facts. I don't think it will affect the provisioning process; it is just something to consider.

ShimShtein: Maybe we should filter out local prefixes like `fe80`?

ShimShtein (commented):
One comment, but otherwise the code looks good.

root/usr/share/ruby/vendor_ruby/discovery.rb
```diff
@@ -231,21 +231,52 @@ def get_mac(interface = 'primary')
...
+  res = {}
+  str = `nmcli -t -f IP6.ADDRESS,IP6.GATEWAY,IP6.DNS con show '#{interface}' 2>/dev/null`
+  return [nil, nil, nil] if $? != 0
+
+  str.each_line do |x|
+    kv = x.split(':')
+    res[kv[0]] = kv[1..].join(':').chomp
+  end
+
+  [res['IP6.ADDRESS[1]'], res['IP6.GATEWAY'], res['IP6.DNS[1]']]
```
ShimShtein, line 278: I think we can filter out the link-local IPs here. Instead of always taking `[1]`, we can iterate through them and pick the first that is not link-local

stejskalleos: Not sure how to do that.

RIght now the check is in the `if cidr6 && !cidr6.link_local?`

ShimShtein: The result that you get out of command is something like this:
```
IP6.ADDRESS[1]:                         2620:1:1:1::1/64
IP6.ADDRESS[2]:                         fe80::1:2:3:4/64
```

You should iterate through all the lines and select only the one that is not link local instead of always taking the `IP6.ADDRESS[1]` field

stejskalleos: oh yeah now I get it. Ok will do.

root/usr/share/ruby/vendor_ruby/discovery.rb
```diff
@@ -231,21 +231,57 @@ def get_mac(interface = 'primary')
...
+  return [nil, nil, nil] if $? != 0
+
+  str.each_line do |x|
+    kv = x.split(':')
+    res[kv[0]] = kv[1..].join(':').chomp
+  end
+
+  # Filter out link-local addresses
+  cidr6 = res6.filter { |k, _v| k.include? 'IP6.ADDRESS' }
+              .delete_if { |_k, v| v.include? 'fe80' }
```
ShimShtein, line 280:
```suggestion
              .delete_if { |_k, v| IPAddr.new(v).link_local? }
```
Not a must, I can live with the `include` version too

stejskalleos:
New version available:
* [fdi-5.0.1-b8446f5.iso](https://fdi-el9.stejskalleos.cz/fdi-5.0.1-b8446f5.iso)
* [fdi-image-5.0.1-b8446f5.tar](https://fdi-el9.stejskalleos.cz/fdi-image-5.0.1-b8446f5.tar)

The remaster tested and should be :green_apple:

shubhamsg199:
Tested the latest iso https://fdi-el9.stejskalleos.cz/fdi-5.0.1-b8446f5.iso and is working as expected.
The kexec is happenning now and host starts provisioning as sson as we hit provision.

PR #172, opened 2025-03-17
Make SYSTEMD_WANTS udev env variable additive

hb0nes:
I ran into the issue of lldp not correctly showing up in Foreman facts for all interfaces.
udev rules turned out to be the culprit.

Fixed, as per https://www.freedesktop.org/software/systemd/man/latest/systemd.device.html:  
`The specified property value should be a space-separated list of valid unit names.`


stejskalleos (approved):
LGTM

stejskalleos:
Thanks @hb0nes

